PR #562, opened 2020-06-19
WIP : lgtm test3

tesshucom:


tesshucom:
LGTM seems to see only the difference.

PR #564, opened 2020-06-19
Fix PMD-blocker

tesshucom:


tesshucom:
It's an interesting point, but let's conclude the PMD with a huge number of modifications first.
Modify PMDs in order of priority, and then fix LGTM.

Issue #561, opened 2020-06-18
Profile where java11/java8 can coexist(LGTM/PMD)


tesshucom:
v109.1.1 and later will include fixes that were not made to avoid differences with Airsonic.
First, clear easy defects of the code.

Somewhat annoying problem with how to operate static tools.

 - Cannot specify branch in LGTM. There is a habit of operation more than Travis.
 - eclipse PMD can run Java8.

tesshucom:
The compiler is changed as follows due to this setting change.

 - java11 is used in travis.
 - java8 is used in LGTM.com.
 - Also, the repository settings have been changed. In the subsequent PR, both Travis and LGTM verification will be performed.
 - In IDE, follow IDE settings. LGTM/PMD plugin can be used in Java 8 environment.
 - Maven depends on environment and profile. To compile using the Java11 boot classpath in the Java11 or later environment, specify the options as follows.

```
 mvn clean package -Dmaven.test.skip=true -Prelease11
 mvn clean package -Dmaven.test.skip=true -Prelease11 -Pjetty-embed 
```

tesshucom:
The lgtm and pmd branches will be created and modified from time to time.
Each will have more commits.
The merge to snapshot is done before release.

tesshucom:
PR to the master in appropriate block units.
(Because LGTM cannot detect it otherwise)

The tesshucom/jpsonic master will continue to import the fix for a while.
These are not immediately reflected in jpsonic / jpsonic, but merged at release.

tesshucom:
Apply a patch that eliminates PMD (blocker/critical) and LGTM (error) to the tesshucom/jpsonic#master.
The compile options will change as covered in this issue.

This patch will not be posted on jpsonic/jpsonic for some time.

tesshucom:
#567
When all the high-priority items were resolved, one new item was detected.

- Resolve newly detected items
- Resolve middle priority. Resolve LGTM before PMD because the amount of modification is huge.

tesshucom:
> There is room for improvement in image cache processing.

The existing bugs with images were not fixed. Should be treated as a separate issue.

tesshucom:
Some issues have occurred and need to be fixed.

 - There is a place where sort is broken
 - For music folder pages, scanning is performed on page transitions

tesshucom:
Low priority issues will be dealt with in other issues.

PR #567, opened 2020-06-20
High priority refactoring (#561)

tesshucom:
A patch that eliminates PMD (blocker / critical) and LGTM (error).


tesshucom:
> 1 for Cross-site scripting

The newly added warning about cross-site scripting is a false positive.
For podcast download streams. No response is taken as it is not a realistic threat.

PR #571, opened 2020-06-21
Fix lgtm wornings

tesshucom:
2 Recommendations and 12 Warnings.


tesshucom:
Some issues have occurred and need to be fixed.

 - There is a place where sort is broken
 - For music folder pages, scanning is performed on page transitions

Issue #586, opened 2020-06-29
Slightly larger refactoring (PMD/Spotbugs)


tesshucom:
- PMD has been fixed for Blocker/Critical/Urgent.
 - Spotbugs has been fixed for High/Medium.

These will be released in v109.2.0.

Some rules are high in number of defects or require design consideration. Therefore, issues are issued individually.  Of these, some may be released in v109.2.0. Others will be released in later versions. No one has ever done that, so I don't have to rush it.

PMD/Spotbugs | rule | comment | sub issues
-- | -- | -- | --
PMD | AccessorMethodGeneration | ~~In most cases it can be safely ignored, but certain patterns can cause problems.~~ unnecessary | #617
PMD | AvoidSynchronizedAtMethodLevel | ~~Must be fixed. Redesign may be necessary.~~ closed  | #591 
PMD | DoNotUseThreads | Must be fixed. Requires a completely different design. | After v109.6.0
PMD | NoInlineStyleInformation | ~~CSS redesign required.~~ closed  | #681
PMD | NoUnsanitizedJSPExpression | ~~Almost simple replacement is expected to work, but it takes time to confirm.~~ closed | #614
Spotbugs | EI_EXPOSE_REP | ~~Easy but large.~~ unnecessary | unnecessary 
Spotbugs | EI_EXPOSE_REP2 | ~~Easy but large.~~ unnecessary  | unnecessary 

higher priority for users : 
DoNotUseThreads [J2EE violation] / AvoidSynchronizedAtMethodLevel [Speed, parallelization, Blocker for future improvements] / NoUnsanitizedJSPExpression [Non-serious security, usability]

Especially Threads needs to be redesigned, so additional design of new Threads should be prohibited until are resolved. Things not covered here, such as exception design, will not be dealt with until they are resolved. Or if the Thread control requires correction of the exception bubbling design, it will be corrected at that time.

- [x] AccessorMethodGeneration
- [x] AvoidSynchronizedAtMethodLevel
- [ ] DoNotUseThreads
- [x] NoInlineStyleInformation
- [x] NoUnsanitizedJSPExpression
- [x] EI_EXPOSE_REP
- [x] EI_EXPOSE_REP2

The following is not required.

##### EI_EXPOSE_REP, EI_EXPOSE_REP2
Performance deteriorates. Since it is a single project, there is no problem if there is no bug.

##### AccessorMethodGeneration 
[...]

PR #592, opened 2020-06-29
Fix urgent issues (#586)

tesshucom:
It's resend of #587.

Add PMD ruleset and modified the source to support priority3.
Some topics require redesign, so they are treated separately.


tesshucom:
Add PMD ruleset and modified the source to support priority3.
Some topics require redesign, so they are treated separately.

PR #612, opened 2020-07-06
Fix supreesion comment

tesshucom:
Check if suppressed comments do the job.


tesshucom:
It seems that the alert has already been deleted from the statistics, while the suppression can be ignored and the alert can be confirmed on the Web screen. not bad. This PR ends. Rewrite the annotation a little and commit directly.

PR #847, opened 2021-01-26
Apply new code style

tesshucom:
Probably one warning will be issued at LGTM


tesshucom:
I thought I would get an error with WebFontUtils, but it seems to have passed. I don't know if it will be detected by master merge.

Issue #977, opened 2021-04-25
Some tuning


tesshucom:
Improves streaming and logging.

____

Items that have different optimal solutions depending on the environment are provided on advanced settings web page.

- [x] PlayQueueInputStream refactoring. In reality, a little over-treatment is woven into it. By rewriting, it is possible to improve the response when the playback position is changed. In other words, at present, the response to changes in streaming conditions is slow. Not only usability for user, but also the responsiveness to the stop command at shutdown is greatly affected. Currently, it is the number one factor in delaying the required processes during shutdown.
- [x] [Advanced settings]: Change the buffer size of sending data. There is a problem of how value to be changed, but the current situation is too small.
- [x] [Advanced settings]: Simplify the log when the server starts.
- [x] [Advanced settings]: Simplified scan log.
- [x] [Advanced settings]: Simplified playing log. Add an option and turn it off by default. Currently, a log is output for each transcoding. It's also a powerful clue when debugging for users, but it can cause the logs to grow significantly during normal operation.
- ~~Suppresses logging when referencing the JNDI name of ThreadFactory.~~ Include it in the log when the server starts. Correct at the same time
- [Advanced settings]: Simplify the log when the server shutdown. -> #829 
- [x] ~~[Advanced settings]:~~ Extension of IdleTimeout for Http. (This is a specification change, not an option. If connected by a specific client app, the player settings will be mass-produced for each number of connections created. This issue should be resolved at the same time.)
- [x] ~~[Advanced settings]:~~ EofException suppression. Suppress rather than optional as it is similar to AbortException
- ~~[Advanced settings]: Database log~~ It will be released at WARN Level from v110.0.0.

____

There is no direct relationship, but it is advisable to start first in order for #829 to proceed smoothly.

Issue #833, opened 2021-01-18
Thread improvements


tesshucom:
Related  #824. Rewrite all the old threaded code.  Divide into several stages.
Release timing will probably be split, as it will be a daunting task. For a while, the tasks are a little fluid.

____

This range should be completed in bulk first.

- [x] Create Executors -> #971
- [x] Rewrite thread : LastFMScrobbler
- [x] Rewrite thread : ListenBrainzScrobbler
- [x] Rewrite thread : AudioPlayer
- [x] Rewrite thread : InputStreamReaderThread
- [x] Rewrite thread : TranscodeInputStream
- [x] Rewrite thread : UPnPService
- [x] Rewrite thread : PodcastService
- [x] Rewrite thread : MediaScannerService
- [x] Graceful shutdown -> #829
- [x] v110.0.0: Fix all pools except scan-pool to fixed size. If the size is exceeded, it will fall back to the pooling of the container, so specify the same value for the queue size and corePoolSize.

#### Transparent function changes for each version

<details>
<summary>v110.0.0</summary>
Ultimately, the user should at least be able to adjust the pooling size of the scan. Scans are pooled in v110.0.0, but are served in the same single thread as before. Thread pooling other than that, there is no big problem with fixed adjustment. 

For the short-pool (for general-purpose short processing), the following values are the initial values in v110.0.0. The advantage of use short-task-pool is that they are easier to profile and the task priority can be lower than the common pool.

short-task-pool.corePoolSize=4
short-task-pool.queueCapacity=4
short-task-pool.maxPoolSize=7

Especially when the user is using it alone, a pool of about up to 4 is sufficient for efficient processing of playing music. If you expect multiple users, you don't need a large pool. Fall back to the common pool (jetty: qtp, tomcat: nio) as before.

</details>

v110.1.0 There is nothing special in this version.

_____

From here on, it's a good idea to separate the release versions.

<details>
<summary>v11x.x.x</summary>

##### Dynamic changes in scalability

If anything, the system must always ensure playing and scanning. Other than that, it can be postponed or rejected. We may need derivative features such as adjusting the scalability of the Executor during the scan or disabling certain web page. If you increase the number of parallelization points other than the thread processing that already exists, the number of processes that can be executed at the same time will increase. Adjustment design is required.

##### Subdivision of scanning process

[...]

Issue #829, opened 2021-01-13
Graceful shutdown


tesshucom:
The current shutdown is based on the Inherently good. Normal operation is unlikely to cause major problems (some persist after VM shutdown), but it allows you to free threads and resources more carefully.
Implement proper shutdown to prevent resource locks in use. Also, shut down as quickly as possible while ensuring the condition that the data can be recovered by rescanning after the next startup.

___

 - Set the shutdown timeout of Spring Graceful to 1 second. There is usually a grace period for user on web application servers shutdown, but this system does not require it. Also, if you set a long grace period, it will wait the same amount of time (if the user keeps the connection, the 503 response to the connected connection will retry. At Jetty.). A shutdown timeout of 1 second simplifies the design of waiting for subsequent pooling discards.
 - Many processes that require closing resouces can be categorized as stream / juke / podcast / scanning, and a discard process must be implemented. In addition, by executing these processes in pooling, it is possible to wait and control for the end of the discard process or to finally forcibly discard due to the time limit.
 - Stops the network layer immediately, but all thread pooling awaits discard threads. 
 - Many shutdown operations rely on the database and cache. If they are destroyed first, they cannot be shut down gracefully. All thread pooling needs control resource dependency with `@DependOn`. Database shutdown is the last.
 - Cache closing used the default listener in legacy implementations, but you need your own disposer to control the closing order.
 - Discards of pooling await up to 30 seconds. In most cases, the destruction of each logic ends first. It is usually possible to shut down immediately.
 - Podcast downloads and scans can take more than 30 seconds. IO speed is relevant. (Operation cases that are unthinkable in normal operation, except for power outages.) Or a directory that stores a huge number.
 - ------> If the podcast download does not successfully update its state due to a shutdown, the episode will be discarded at the next boot. (Existing specifications)
 - ------> CHECKPOINT is not executed if shutdown occurs during scanning. This is a specification change by Jpsonic. in the case of a legacy server, since there is no shutdown process, it will be flushed without executing CHECKPOINT. (The implementation is different, but the operation of CHECKPOINT is the same as a result) The difference is where to break the process.

The scan processes up to the directory in the process and breaks loop. Later versions may change to finer cancellation controls.

___

 - [x] Enable Graceful shutdown of Spring
 - [x] stream
 - [x] juke
 - [x] podcast、scanning
 - [x] scanning
 - ExitCodeGenerator(... Probably 130 in most cases)

___

tesshucom:
[...]

PR #990, opened 2021-05-02
Fix for thread and shutdown

tesshucom:
- Thread improvements (#833)
 - Add logging control page, some tuning (#977)
 - Graceful shutdown (#829)


tesshucom:
(Abandoned)

Issue #1349, opened 2022-02-17
Add DSD Support


tesshucom:
- [ ] Add new transcoding commands. Only FLAC conversion will be supported, as conversion to MP3 seems to be in low demand.
 - [ ] Add MIME. Since it is not a general MIME, it needs to be selectable unlike other formats.

PR #1351, opened 2022-02-19
Add DSD Support (#1349)

tesshucom:


tesshucom:
Re-reviewed on # 1352

PR #1495, opened 2022-05-06
Fix null dereference

tesshucom:
- The null dereference pointed out by Sonatype is resolved.
 - The null dereference pointed out by Sonatype is included in the spotbugs NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE check item. 
   - The purpose of this pull request is to eliminate alerts about this.
   - Therefore, a wider range of corrections will be made than Sonatype's advice.
____

In most cases, the cause of the correction is one of the following.

 - When null occurs in the process of creating a user object from HttpRequest. Access by an unauthorized user, or when the user account is deleted while browsing, etc.
 - Illegal path processing. In most cases, the path object handled by this system is NonNull. However, for directory roots, certain operations will result in null (Including the case where parent indirectly refers to the route by the path operation). In legacy code, it was NPE in such cases.

So it's a fix for null references caused by runtime. This fix changes to explicitly throw an IllegalArgumentException instead of an NPE in such cases.

eg.

> throw new IllegalArgumentException("User not found:" + username)
> throw new IllegalArgumentException("Illegal path specified: " + path)
____

As a result, **nothing changes in the behavioral specifications** from the user's perspective. However, unlike the NPE exception message, we can determine if it is a trivial unchecked exception.

### DIFF
--- a/jpsonic-main/src/main/java/com/tesshu/jpsonic/service/PodcastService.java
+++ b/jpsonic-main/src/main/java/com/tesshu/jpsonic/service/PodcastService.java
@@ -355,6 +355,9 @@ public class PodcastService {
 
             File dir = getChannelDirectory(channel);
             MediaFile channelMediaFile = mediaFileService.getMediaFile(dir.toPath());
+            if (channelMediaFile == null) {
+                return;
+            }
             Path existingCoverArt = mediaFileService.getCoverArt(channelMediaFile);
             boolean imageFileExists = existingCoverArt != null
                     && mediaFileService.getMediaFile(existingCoverArt) == null;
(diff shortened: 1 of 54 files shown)


jpsonic-main/src/main/java/com/tesshu/jpsonic/service/PodcastService.java
```diff
@@ -355,6 +355,9 @@ private void downloadImage(PodcastChannel channel) {
 
             File dir = getChannelDirectory(channel);
             MediaFile channelMediaFile = mediaFileService.getMediaFile(dir.toPath());
+            if (channelMediaFile == null) {
+                return;
+            }
```
tesshucom, line 360: If channelMediaFile is null, no further processing will be performed.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/PodcastService.java
```diff
@@ -736,8 +739,10 @@ private File getChannelDirectory(PodcastChannel channel) {
             }
 
             MediaFile mediaFile = mediaFileService.getMediaFile(channelDir.toPath());
-            mediaFile.setComment(channel.getDescription());
-            mediaFileService.updateMediaFile(mediaFile);
+            if (mediaFile != null) {
+                mediaFile.setComment(channel.getDescription());
+                mediaFileService.updateMediaFile(mediaFile);
+            }
```
tesshucom, line 745: Comment update is executed only when mediaFile can be obtained

Issue #1539, opened 2022-06-05
UploadController fixes


tesshucom:
Tiny fixes for UploadController.

### Problem description

 - PMD.UseIOStreamsWithApacheCommonsFileItem
   - > Commons-fileupload is used for uploading. FileItem.get() and FileItem.getString() could exhaust memory since they load the entire file into memory. 
     - However, false positives. PMD's claim is correct on the API. However, in the implementation of commons, FileItem.getString () of the interface behaves differently depending on the data variation of the instance. The risk case of overflow is when the instance is a file. Therefore, there is no problem in implementation. (The problem is the API design of commons-file upload)
     - It had been a complex issue and was suppressed by ruleset.xml. Since there is only one applicable place, it will be changed to suppression by annotation
 - Not a process we want to do in parallel with the scan. It is hoped that it will be suppressed during the scan. Related #1475.
 - Degradation fix for "uploaded files"
   - Commits that have occurred : e342496, [comment1](https://github.com/tesshucom/jpsonic/commit/e3424964944c04fde65ec11c22800250267cb862?diff=split#r75368404), [comment2](https://github.com/tesshucom/jpsonic/commit/e3424964944c04fde65ec11c22800250267cb862?diff=split#r75368222)
   - It's not fatal. In the case of a regular file upload, not a Zip file. The process works fine, but there is an probrem in creating the list of uploaded files used for notifications.
   - This degradation can be detected in UnitTest. [UploadControllerTest](https://github.com/tesshucom/jpsonic/blob/master/jpsonic-main/src/test/java/com/tesshu/jpsonic/controller/UploadControllerTest.java). However, since it is not possible to execute multiple test cases now, the corresponding test case was commented out. :kissing:
     - Test case is not specifically changed. There are many cases where updating Junit will solve the problem naturally.

### Fixes

 - [x] PMD(urgent) : UseIOStreamsWithApacheCommonsFileItem (suppression by annotation)
 - [x] Add restriction will be placed on uploading during scanning
 - [x] Degradation fix for "uploaded files"

PR #1542, opened 2022-06-07
UploadController improvements (#1539)

tesshucom:


jpsonic-main/src/main/java/com/tesshu/jpsonic/controller/UploadController.java
```diff
@@ -143,13 +149,20 @@ protected ModelAndView handleRequestInternal(HttpServletRequest request, HttpSer
...
+            String contentDisposition = item.getHeaders().getHeader(FILE_ITEM_HEADER_DISPOSITION);
+            if (contentDisposition != null) {
+                Matcher m = fileItemHeaderValue.matcher(contentDisposition);
+                if (m.find()) {
+                    return Path.of(m.group(1));
+                }
             }
         }
-        return null;
+        throw new IOException("Missing 'dir' parameter.");
```
tesshucom, line 165: It works in unit tests, but may not work when integrated ... ?

tesshucom, line 165: It works in unit tests, but may not work when integrated ... ?

tesshucom:
This pull request will be rejected. There are some complicated issues. So organized by #1539.

PR #1600, opened 2022-07-12
Fix bug that UPnP port specified is not reflected correctly

tesshucom:
Degrading at 24d54fd.

### DIFF
--- a/jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/ApacheUpnpServiceConfiguration.java
+++ b/jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/ApacheUpnpServiceConfiguration.java
@@ -23,6 +23,7 @@ package com.tesshu.jpsonic.service.upnp;
 
 import java.util.concurrent.Executors;
 
+import com.tesshu.jpsonic.service.SettingsService;
 import org.fourthline.cling.DefaultUpnpServiceConfiguration;
 import org.fourthline.cling.transport.impl.apache.StreamClientConfigurationImpl;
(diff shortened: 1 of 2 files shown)


jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/ApacheUpnpServiceConfiguration.java
```diff
@@ -48,6 +49,8 @@ public StreamClient createStreamClient() {
     @SuppressWarnings("rawtypes")
     @Override
     public StreamServer createStreamServer(NetworkAddressFactory networkAddressFactory) {
-        return new StreamServerImpl(new StreamServerConfigurationImpl(networkAddressFactory.getStreamListenPort()));
+        int listenPort = SettingsService.getDefaultUPnPPort();
+        return new StreamServerImpl(new StreamServerConfigurationImpl(
+                listenPort < 1 ? networkAddressFactory.getStreamListenPort() : listenPort));
     }
```
tesshucom, line 55: Jpsonic's UPnP port defaults to -1. (Represents unspecified). If the UPnP port number specified by the user is less than 1, Jpsonic's UPnP port determination has the same specifications as Subsonic. In other words, Auto detect is used.

If the value is above 1, then use that port value to launch UPnP.

tesshucom: This specification is spec of Jpsonic, but the fix of https://github.com/tesshucom/jpsonic/commit/24d54fd36d5f89e93cc0ade31d14b9959da7cb73 was not enough and caused a bug.

Issue #839, opened 2021-01-20
Bump Tomcat && Jetty && Spring Boot


tesshucom:
Prerequisites: #2305 

### Problem description

A major version upgrade of Jetty will be done in v111.3.0. #838 On the other hand, Tomcat will maintain the status quo. 

___

The comparison of each product version is as follows. The impact of Jakarta EE 9 is great, but now it is just before that. We need to keep an eye on new information.

Jpsonic | Jetty | Tomcat | Servlet | EE / Spring Boot | Namespace |  
-- | -- | -- | -- | -- | -- | --
111.2.0 | 9.4.x | **9.0.x**  | 3.1 | Java EE 7 / 2.x | javax.servlet | 8 and later
**111.3.0** | **10.x** | **9.0.x** | **4** | **Jakarta EE 8 / 2.x** | **javax.servlet** | **11 and later**
? | 11.x | 10.0.x | 5 | Jakarta EE 9 / 3 | jakarta.servlet |  
? | 12.x? | 10.1.x | 6 | Jakarta EE 10 |   |  

### Additional notes

 - [Preparing for Spring Boot 3.0](https://spring.io/blog/2022/05/24/preparing-for-spring-boot-3-0)
   - Upgrade to Java 17 :white_check_mark:
   - Upgrade to the Latest Spring Boot 2.7.x :white_check_mark:
   - Check for Calls to Deprecated Code. ``All deprecated code will be removed.`` :heavy_exclamation_mark: 
   - PathPatternParser (Recommendation) :kissing:
   - Check if Third-party Projects Have Jakarta EE 9 Compatible Releases :worried:
     - Tomcat and Jetty should be upgraded at the same time
     - [Upgrading Tomcat 10.0.x](https://tomcat.apache.org/migration-10.html#Upgrading_10.0.x)
     - https://github.com/directwebremoting/dwr/issues/39
     - #1630
       - It can be avoided by downgrading to DWR 3.0.0-rc3-RELEASE and seems to work with Java17. Worst case, there will be a way to not upgrade DWR until Java 11 is EOL. Premier Support of Java 11 by Oracle Expires September 2023
   - Try the Spring Boot Milestones :expressionless:
     - > Although we don’t recommend it for production, you can try Spring Boot 3.0 milestones today to see how hard it will be to migrate your project.  

___

There are some barriers to resolving this issue. However, there is a grace period until it is completely resolved.

[...]

PR #1602, opened 2022-07-13
Change shutdown timeout

tesshucom:
### Problem description

Change shutdown timeout to 29 seconds. This is because the timeout phase timeout value, including Docker, is fix.

 - In a general environment, it is often set to about 30 seconds.
 - For this reason, important processing had a grace period of about 30 seconds. (Many threading await times had been set to 30 seconds)
 - This fix will change the entire shutdown phase to 29 seconds (Many threading await times will be set to 25 seconds)

### Additional notes

 - In Docker this value is managed by stop_grace_period, which is 10 seconds by default. A light weight shutdown. That's good, but 10 seconds is rough.
 - Jpdonic assumes that stop_grace_period is 30s.
 - The settings related to the number of seconds should be as follows:
   - **Docker : stop_grace_period**  > **Spring : spring.lifecycle.timeout-per-shutdown-phase**  > **The await termination value of ThreadPool**
   - ExecutorConfiguration manages threads with long processing time. We can protect from a violent shutdown, by setting the appropriate '@DependOn' for these Pooling, cache, and 'liquibase' been.

This is the current workaround.

### DIFF
--- a/jpsonic-main/src/main/java/com/tesshu/jpsonic/util/concurrent/ExecutorConfiguration.java
+++ b/jpsonic-main/src/main/java/com/tesshu/jpsonic/util/concurrent/ExecutorConfiguration.java
@@ -42,10 +42,10 @@ public class ExecutorConfiguration {
 
     private static final Logger LOG = LoggerFactory.getLogger(ExecutorConfiguration.class);
 
-    protected static final int SHORT_AWAIT_TERMINATION = 20_000;
-    protected static final int PODCAST_DOWNLOAD_AWAIT_TERMINATION = 30_000;
-    protected static final int PODCAST_REFRESH_AWAIT_TERMINATION = 30_000;
-    protected static final int SCAN_AWAIT_TERMINATION = 30_000;
+    protected static final int SHORT_AWAIT_TERMINATION = 10_000;
+    protected static final int PODCAST_DOWNLOAD_AWAIT_TERMINATION = 25_000;
+    protected static final int PODCAST_REFRESH_AWAIT_TERMINATION = 25_000;
+    protected static final int SCAN_AWAIT_TERMINATION = 25_000;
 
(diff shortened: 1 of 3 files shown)


tesshucom:
The mechanism of Spring's graceful shutdown and the mechanism of exit code are closely related. There are some traps, and the behavior depends on the version of Spring Boot. Therefore, fragmentary changes can cause deadlocks. Best practices do not appear to be finalized. 

What is being done with Jpsonic is a workaround that assumes a shutdown hook or a graceful-shutdown and can work with the latest Spring. Spring Boot versions are not stagnant or reverted for a particular implementation. The old Spring Boot implementation method doesn't help. The #839 needs to be resolved in the end, so it should always be checked with the latest version.

``https://github.com/spring-projects/spring-boot/issues/27049``
``https://github.com/spring-projects/spring-boot/issues/23625``

Explicit control of the shutdown code is currently difficult. Therefore, Jpsonic will reply with the received signal as it is. (That is 143, for Docker) 

> If the exit code mapping by exception does not cause a deadlock, there may be a tricky workaround to take advantage of it...  In other words, it seems better not to uselessly implement any more. If Spring Boot provides good practices in the future, you should follow them.

Exit code issues are not related to issues such as:

``https://hynek.me/articles/docker-signals/``

Message transmission from Docker to Java is successful. (We can receive it, but the latest Spring Boot does not allow good coexistence of exit code control and shutdown process.)

Therefore, at present, it is a realistic idea to map the exit code with tini as follows.


> ENTRYPOINT ["tini", "-e", "143", "--", "run.sh"]

It may be better to maintain the above method for some time.

___


 - In SpringBoot, the shutdown hook, context close, and exit code control conflict now. If this is resolved, better implementation may be possible
 - It seems that a standard shutdown implementation is being considered in Liquibase as well. This may also be positively affected in the future

With these in mind, realistic workarounds should be accepted.

PR #1636, opened 2022-08-04
Change specification of font

tesshucom:
### (1/2)Change specification of font used for embedded cover art

The font specification has been changed to use the Java logical font by default. If you want to use Japanese fonts like before, use the option below.

`-Djpsonic.embeddedfont=true`

For Docker, change the following items in production.yml.

`EMBEDDED_FONT: true`

#### Background

The reason why Jpsonic had been used embedded fonts by default is as follows

 - Tofu happens when the JDK is not configured to use the font system correctly. Many Japanese will see tofu on Linux. Of course it can be avoided by properly configuring fonts, but it may not be easy.
 - JDK transition from Oracle to OpenJDK could caused font problems. 
 - Rendering with embedded fonts is possible even on systems where fonts do not exist. Insurance feature.

The environment surrounding fonts is a little different than it was 5 or 10 years ago. The current Jpsonic can render like the following with Docker's initial configuration. No Tofu !

image

#### (2/2)Fix ffmpeg/ffprobe links

Fixed ffmpeg/ffprobe links.

 - Current ffmpeg latest is 5.1.
 - `ffmpeg -version`
 - Alpine uses ffmpeg 5.0.1. Almost the same as the latest version.
 - Ubuntu uses ffmpeg 4.x. It's a little old, but the composition is almost the same
 - `ffmpeg -formats`
 - If we grep ffmpeg supported formats with Jpsonic's standard format, the results are the same between Alpine and Ubuntu.

By compiling ffmpeg using multi-staging, more formats can be supported. However, it takes several tens of minutes, so I would like to avoid it if possible. FFmpeg lib binaries for Alpine and Ubuntu are available from their respective repositories. On AWS (or Cent family),  we are forced to compile ffmpeg.

### DIFF
--- a/install/docker/alpine-jdk17/Dockerfile
+++ b/install/docker/alpine-jdk17/Dockerfile
@@ -13,6 +13,10 @@ ENV JPSONIC_DIR=/jpsonic \
     LANG=C.UTF-8 \
     TIME_ZONE=GB \
     BANNER_MODE=on \
+    SCAN_ON_BOOT=false \
+    EMBEDDED_FONT=false \
+    MIME_DSF=audio/x-dsd \
+    MIME_DFF=audio/x-dsd \
     JAVA_OPTS=-Xmx512m
 
 WORKDIR $JPSONIC_DIR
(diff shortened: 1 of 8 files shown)


tesshucom:
There seems to be a problem with armv7/mstl configuration. This pull request will be reissued.

PR #1654, opened 2022-08-19
Change upstream of all Docker images to Temurin

tesshucom:
In v111.5.0, all upstream images will be unified to Temurin, and the beta version will end. 

The image tag matrix looks like this:

<table cellspacing="0" cellpadding="0" border="1">
<tr>
<th>TAG</th>
<th>JAVA</th>
<th>OS:ARCH</th>
</tr>
<tr>
<td>[vxxx.x.x, latest]</td>
<td>17-jre</td>
<td class="s3" dir="ltr" rowspan="2">Alpine:amd64</td>
</tr>
<tr>
<td>ea</td>
<td>17-jdk</td>
</tr>
<tr>
<td>[vxxx.x.x, latest]-jammy</td>
<td>17-jre</td>
<td class="s3" dir="ltr" rowspan="2">Ubuntu: amd64, arm64, arm/v7</td>
</tr>
<tr>
<td>ea-jammy</td>
<td>17-jdk</td>
</tr>
</table>

And tests will be added and minor fixes to production.yml will be made.

#### Fonts

Temurin uses Dejavu as the default font. Strictly speaking, Alpine and Ubuntu have different font configurations (Alpine probably has better coverage). In Jpsonic add Noto CJK to these and change to default. 

 - In general, Noto is more popular
 - Music titles and artists may use difficult kanji characters that are not often used. Noto CJK covers Chinese characters used in Taiwan, Hong Kong, China, and Japan. The original developer is Adobe and is highly reliable, and beautiful.
 - Even if you add other Noto, the affinity is high (such as adjusting the size). 
 - Noto, which is newer than Dejavu, is expected to have better coverage and future maintenance.


tesshucom:
Disappointing. Failed to map. local.conf is incorrect, and should be sans-serif instead of serif. This pull request will be republished.

install/docker/alpine-jdk17/local.conf
```diff
@@ -0,0 +1,12 @@
+<?xml version='1.0'?>
+<!DOCTYPE fontconfig SYSTEM 'fonts.dtd'>
+<fontconfig>
+    <match target="pattern">
+        <test qual="any" name="family">
+            <string>serif</string>
```
tesshucom, line 6: Family is wrong. [sans serif/sans]

PR #1681, opened 2022-09-12
LGTM/CodeQL : Overly permissive regular expression range

tesshucom:
Suppression of false positives.

 - False positive for https://lgtm.com/rules/1515921952940/
 - Because the input is one character

As usual, the alert pop comes LGTM first. Luckily, CodeQL on Github also generated a similar warning a few days later.

___

LGTM will be going out of service.

> LGTM.com will be shut down in December 2022 — we recommend that you use GitHub code scanning instead. Read more in [our blog post on the GitHub blog](https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/).

Therefore, the LGTM badge should be removed and CodeQL required for workflows. It seems to be running for a while, so activate both and observe the difference in behavior.


tesshucom:
Suppression of LGTM does not take effect immediately, so merge and see how it goes

PR #1688, opened 2022-09-17
Add ZoneOffset to Internal help page

tesshucom:
Related #1680.

``About > Internal Details > Locales`` changes.

 - Change item name. Change some item names to show what they are outputting
 - Add and Show the time zone recognized by Jpsonic

___

- [ ] After merging, check the output in Docker (Alpine/Ubuntu Jammy) and get a snapshot
- [x] Get snapshot on JDK18. Related [JEP 400](https://openjdk.org/jeps/400).


tesshucom:
file.encoding has been alrerdy "deadly" . The important thing is the default character set. In Jpsonic, Charset.defaultCharset() is explicitly set in all readers. In the future, we will also consider supporting these change options.

> java -version
> openjdk version "17" 2021-09-14
> OpenJDK Runtime Environment (build 17+35-2724)
> OpenJDK 64-Bit Server VM (build 17+35-2724, mixed mode, sharing)

image

> java -version
> openjdk version "18" 2022-03-22
> OpenJDK Runtime Environment (build 18+36-2087)
> OpenJDK 64-Bit Server VM (build 18+36-2087, mixed mode, sharing)

image

tesshucom:
#### Alpine

image

#### Ubuntu Jammy

image

tesshucom:
There is no problem at present, but it seems that the JP file of i18n is not corrected. (The contents are almost the same, but this pull request will be discarded and updated)

PR #1888, opened 2022-11-16
Add Jsonp to Suppressed legacy features

tesshucom:
Pointed out by JGTM.

> Writing user input directly to a web page allows for a cross-site scripting vulnerability.

It's to write callbacks if Jsonp is used in Subsonic API.

 - The API is not intended for use by any third party. So it's not an immediate threat.
   - Even in Subsonic and Airsonic, Jsonp is not executed unless token authentication is passed
 - But it would be an anachronism to have Jsonp enabled by default... Disabling the default doesn't bother anyone.

Therefore, Jpsonic is modified so that Jsonp cannot be used by default. Only available if explicitly enabled by an administrator in Settins > General > Suppressed legacy features.

image


tesshucom:
The following two LGTM alerts are suppressed. All already fixed (false positives).

 - https://lgtm.com/rules/3960079/
 - https://lgtm.com/rules/5970070/

LGTM will not be available at the end of this month, so this will probably be the last fix. Its successor, CodeQL, is already integrated and used in CI.

[The next step for LGTM : GitHub code scanning!](https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/)

The detection accuracy is almost the same, but since CodeQL suppresse false positives with the web page, there is no suppression history on the source code. (There are good and bad sides...)

PR #1896, opened 2022-11-17
Bug fix for welcome page message

tesshucom:
The following bugs will be fixed.

 - Link not available in French due to bad tag (from Airsonic)
 - Malfunction due to double quotes in Script. Lack of the escaping. (from Airsonic).
 - Link not available in all languages except Japanese and Brazilian due to bad tags (from Jpsonic)

Also, as an extension, we will add the ability to switch the default language of the server, to the welcome page.


tesshucom:
image

PR #1936, opened 2022-12-01
Split MediaFileService into R and CUD

tesshucom:
Prerequisites: #1922

Fixed MediaFileService to be limited to readOnly methods. Move methods that may write to the database into a new class called WritableMediaFileService. 

Besides the simple method move, the following fixes are also included:

 - The method that used to do all CRUD in one method will be split into two, one dedicated to R and the other.
 - "Create or Insert Methods" are originally divided into those that are required and those that must not be performed. Methods that should not do so will have new "Update methods" created.
    - Updates using createOrUpdate and physical deletes may conflict
    - Executing on Java side select -> createOrUpdate on the same record outside of a transaction will lead to conflicts and unintended overwrites depending on the timing.
    - Hard to happen. Of course. However, even if it is a glitch with a low probability of occurrence, the probability of occurrence may increase if we continue to make improvements to shorten the time required for the same processing.

There are 3 main aspects of the fix in this pull request.

#### Encapsulating CRUD processing for tables used for scanning

 - WritableMediaFileService is located in the service.scanner package. As long as these methods are package-private, they cannot be used by other packages.
 - In #1922, the service.scanner package implements strong locking to prevent similar long-running batches from running concurrently to prevent scan conflicts. Basically it is assumed that these classes use WritableMediaFileService. In other words, the design was changed to be able to protect against update conflicts.
 - Design changes needed to support reliable cancellations, retries, and parallel execution.

image

#### Conflict-free database registration

Not all database entries can be rejected during scanning. For example, incrementing the number of performances of the number of songs must operate without conflict even during scanning. There are other processes with similar characteristics, such as editing tags and comments, and changing image paths. 

 - These methods have been changed to be provided as public methods in WritableMediaFileService.
 - All public methods present in WritableMediaFileService are fixed to work without conflicts with scanning.

#### Remove index update function

Since Subsonic, there are some functions for updating the view during scanning.

 - **Fast access mode** : When this is turned off, if the data does not exist in the DB, refer to the actual file, parse it in real time, and return the result to the view. It's a flawed feature that can trigger a memory overflow under certain conditions. Removed in Jpsonic already. 
 - **Refresh button** : This is likewise a feature that may trigger parsing in real time. Even when it's scanning. In Jpsonic, this has been added as a "Suppressed Legacy Features" and changed to be disabled by default. Jpsonic plans to remove the Refresh Button in #1937.

This pull request will remove the internal implementation involved in the refresh button.


tesshucom:
It is currently under review and will not be completed soon. In addition to MediaFiles, Albums and Artists are also scrutinized (probably a small number)

tesshucom:
This pull request reduces the usage of createOrUpdate. This makes future redesigns much easier.

MediaFileDao#createOrUpdateMediaFile (5)
ArtistDao#createOrUpdateArtist(1)
AlbumDao#createOrUpdateAlbum (1)

Also, due to design changes in #1922 and this pull request, the analysis of sonatype-lift changes as follows. Appropriately correcting the design seems to output useful information to some extent.

<details>
<summary>According to the sonatype-lift analysis</summary>

> #### Refactor God Classes Graphical Overview
> This chart suggests which God Class you may wish to fix first, based on the class's Change Proneness vs Effort to Fix. Classes on the top left of the graph are considered higher priority, because they are edited often and are easier to refactor.
> image

</details>

The upper right class in the figure (high effort) is the class originally expected to be fixed in v111.6.0 and v111.7.0. It's hard to fix, but the return we get is big.

PR #1951, opened 2022-12-08
Minor fixes in workflow 

tesshucom:
Fixes to resolve Annotations that recently started appearing in Github Actions.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -224,11 +224,11 @@ jobs:
         run: cp install/keys/README.txt jpsonic-main/target
       - name: Get current date
         id: date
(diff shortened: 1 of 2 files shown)


tesshucom:
#### Why is the above error

Because the Java's default Executor was used in the unit test. Jpsonic executor does not output error with System.err when an exception occurs in an internal thread. (Logger will be used)

When unit testing of exception case by mock is done, it is possible to check the operation even with Java's default executor. However, Github Actions seems to supplement the output by System.err and list it in Annotations.

#### Why is the above warning

In Github Actions commands, save-state and set-output are scheduled to be abolished to improve security.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Issue #1707, opened 2022-09-23
Is image low resolution?


tesshucom:
## Problem description

It looks like the resolution specification is being ignored.

![image](https://user-images.githubusercontent.com/27724847/192014678-f4739080-80be-4e39-8bc2-4aebbda13e9b.png)

### Steps to reproduce

Delete all image caches and regenerate

## System information

 * **Jpsonic version**: v111.5.0
 * **Operating system**: All
 * **Tomcat/standalone**: All

tesshucom:
Include the fix in a hotfix as it is an obvious regression.

tesshucom:
Ha... WMP was the culprit. When using Widows Media Player to check the operation of UPnP, it seems that a hidden file AlbumArtSmall.jpg was created in all directories. After deleting all AlbumArtSmall.jpg, it was restored.

![image](https://user-images.githubusercontent.com/27724847/192134158-f65134a1-ddd3-4527-bc95-12d8c40e88ac.png)

We may have an idea to scan without AlbumArtSmall.jpg. However, it's probably best to do nothing because it masks this bad habit.

tesshucom:
Well ... we can't miss this one. What is bad is software that creates files without permission, but protection measures are necessary.

Issue #408, opened 2020-03-08
Add tag consistency check


tesshucom:
Consider whether a checker page for missing tags can be created.

There are many check items.
Extract from the washing process.

tesshucom:
Do not start immediately.
For that reason, it is removed from the milestone of v107.3.0.

For a while, attach a check that may be a candidate for this issue.

tesshucom:
Implementation as a tool is not urgent.
Now that it's easier to run on a SQL client, I'll publish an introduction to the connection process and some query examples on the Web.

PR #1952, opened 2022-12-08
Add option to exclude specific cover art files

tesshucom:
### Summary

A new option will be added to Settings > General. 

If any image exists in the directory, the file specified here will not be recognized as cover art.

 - `Cover art files` are "suffix match" "ignore case"
 - `Cover art to exclude` are "suffix match" "ignore case"
 - `Cover art to exclude` rule takes precedence

image

### Use case

Jpsonic can use WMP(Windows Media Player) via UPnP. It also supports Composer and multi-genre.

image

image

Unfortunately, WMP creates image caches in the Music folder by default. Moreover, it is a hidden attribute file. In this case, an image for the cache for icon will be created with a lower resolution than the original embedded image. When Jpsonic scans in this state, Jpsonic recognizes WMP's cache as cover art.

Users will (I'll) be confused if they don't know this WMP specification. See #1707.

 - Common story. When using multiple apps and a media server together.
 - Could it be a media server bug? Hmm. If anything, I think it's crazy to have an app that creates a cache in your music folder without your permission.  😡 😡 😡 😡

### I'm already in the swamp!!

Do not worry. Recovery is possible.

 - Note the initial setting for Cover art to exclude : `folder.jpg AlbumArtSmall.jpg small.jpg large.jpg`
   - WMP poops : folder.jpg AlbumArtSmall.jpg
   - I have checked the operation with various applications. I saw different poops. Filtering out anything ending with small.jpg or large.jpg seems cleaner.
   - If this rule is exceeded, please add a new one
 - Please rescan.
   - However, since the file has not changed at all and only the data has to be modified, the option is necessary. Enabling `Ignore timestamps` makes this possible. Don't forget to deactivate it after scanning.

image

It may not be reflected immediately. Ctrl + F5 will probably fix it. If your browser's cache is messed up, it should probably be invalidated after 10 hours.


tesshucom:
#### I want to confirm!!

Certainly. Apart from the scanning function, I would like to create a verification function that can check the integrity of the tag library. (Please fill in if you have a request for a check function)

 - #408

It's just that we're not at that stage yet.

___

At this stage, it is possible to check with SQL and visually.

``
select cover_art_path
from media_file
where cover_art_path notnull;
``

 - If any music file is registered in cover_art_path, it means a candidate for embedded cover art.
   - Jpsonic differs from other Sonic servers in this way. 
     - Only candidate selections of external cover art files or embedded files are included when scanning.
     - When actually using images, more detailed checks for embedded images than other Sonic servers will be performed. A mysterious error was sometimes spewed out on legacy server, but that is not the case with Jpsonic.
     - Legacy servers did these during scans, which contributed to bottlenecks
 - Be careful if any image paths are registered.
   - If it's the intended file, that would be the expected behavior.
   - If an unknown file is registered, some software may have done something wrong. Consider additional settings to Cover art to exclude.
 - Well, maybe. . . I think we need to add some kind of validation feature.

#### I want to clear the trash!

It's impossible to add that functionality. It is possible that the design is as guilty as the feature to create garbage without permission.

In the case of Windows, if music software does something with hidden attributes, it may be troublesome to deal with standard functions such as Explorer. In such cases, you may need a slightly more powerful filer such as [voidtools](https://www.voidtools.com/).

Issue #1986, opened 2023-01-05
Unknown: java.lang.IllegalArgumentException: Comparison method violates its general contract! when browsing.


DutchComputerKid:
## Problem description

Now this is a wierd one, mostly when trying to log in or out this pops up, but from what I remember sometimes with browsing too.
This has been popping up for a while now but since I have not found anything concrete about how to trigger it, I just saved a couple error 'logs'.
[error3.txt](https://github.com/tesshucom/jpsonic/files/10354492/error3.txt)
[error2.txt](https://github.com/tesshucom/jpsonic/files/10354493/error2.txt)
[error.txt](https://github.com/tesshucom/jpsonic/files/10354494/error.txt)

### Steps to reproduce

I'm not certain, like I said it mostly happens when logging in from a normal user to admin user, since as I put in issue #1985, starting JPSonic to do a library scan requires logging into the admin user for now.

## System information

 * **Jpsonic version**: *  Jpsonic 111.5.13 – 2022-12-25*
 * **Operating system**: *Debian 11.6
 * **Java version**: *openjdk 17.0.4 2022-07-19*
 * **Proxy server**: None
 * **Client**: *Firefox, Edge, Chrome*
 * **Language**: *English (United States)*
 * **Tomcat/standalone**: *Docker latest*

tesshucom:
Thank you for your report. I saw it for the first time. But it's the kind of problem that should be fixed. It may take some time to reproduce, but it will be addressed.

tesshucom:
Hi.

We are investigating this issue. From the error message reported, the cause is relatively clear. 

<details>
<summary>the cause</summary>

Sort logic was scattered in Subsonic and Airsonic. Moreover, the content differs depending on the implementation location. In Jpsonic, the sorting logic is managed centrally in one place. But ...

Part of [the Comparator](https://github.com/tesshucom/jpsonic/blob/master/jpsonic-main/src/main/java/com/tesshu/jpsonic/domain/JpMediaFileComparator.java)  has a problem. Sorting logic has two main points of view: Asymmetry and Transitive. Maybe ... the Transitive isn't perfect now.

#### Asymmetry

> Implementers must ensure that signum(compare(x, y)) == -signum(compare(y, x)) for all x and y.

#### Transitive

> Implementers must also ensure that the relation is transitive: ((compare(x, y)>0) && (compare(y, z)>0)) means that compare(x, z)> means 0.

[...]

PR #2014, opened 2023-01-31
Fix the comparison method violates general contract

tesshucom:
## Problem description

Fix for #1986. 

When there is data with a combination of specific data patterns, an error occurs at the time of sorting with probability. Incomplete Transitivity Implementation of Comparator Causes Timsort to Error.

## System information

 * **Java version**: All
 * **Jpsonic version**: Mostly since v111.5.11 (with postgres?)
 
 - This is a latent bug and has been around for a long time. But this is my first bug report.
 - Realistically, it is assumed that the probability of occurrence has increased after v111.5.11. Bug fixed in v111.5.11, Postgres is now available. #1950
   - It's tedious to reproduce situations where sort transitivity violates the convention. Depending on the size of the list and the position of the cause, even if the underlying data pattern is present in the list, it cannot be reproduce. 
     - In my environment, the exact same library was newly scanned with the default embedded DB and Postgres and an attempt was made to reproduce it. I couldn't reproduce it with the default HSQLDB, but I could reproduce it with Postgres by accident.
     - I've never seen it happen in HSQLDB. Combining the sorting implementation before modification with Postgres sorting rules might increase the chances of being reproducible. (HSQLDB and Postgres default sorting rules are different.)

### DIFF
--- a/jpsonic-main/src/main/java/com/tesshu/jpsonic/domain/JpMediaFileComparator.java
+++ b/jpsonic-main/src/main/java/com/tesshu/jpsonic/domain/JpMediaFileComparator.java
@@ -25,6 +25,8 @@ import java.text.Collator;
 import java.util.Comparator;
 
 import org.apache.commons.lang3.ObjectUtils;
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * Comparator for sorting media files.
@@ -64,7 +66,10 @@ class JpMediaFileComparator implements MediaFileComparator {
         }
 
         if (a.isDirectory() && b.isDirectory()) {
-            return compareDirectory(a, b);
+            i = compareDirectory(a, b);
+            if (i != 0) {
+                return i;
+            }
         }
 
         // Compare by disc and track numbers, if present.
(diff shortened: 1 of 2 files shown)


jpsonic-main/src/main/java/com/tesshu/jpsonic/domain/JpMediaFileComparator.java
```diff
@@ -98,21 +104,14 @@ private int compareAlbumAndNotAlbum(MediaFile a, MediaFile b) {
         return 0;
     }
 
-    @SuppressWarnings("PMD.ConfusingTernary") // false positive
-    private int compareDirectory(MediaFile a, MediaFile b) {
-        int n;
+    int compareDirectory(MediaFile a, MediaFile b) {
```
tesshucom, line 107: Comparison between directories. Removed unnecessary comparisons that caused conflicts.

jpsonic-main/src/test/java/com/tesshu/jpsonic/domain/JpMediaFileComparatorTest.java
```diff
@@ -0,0 +1,322 @@
...
+import java.text.Collator;
+import java.util.Locale;
+
+import com.tesshu.jpsonic.domain.MediaFile.MediaType;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
+import org.junit.jupiter.api.Test;
+
+@SuppressWarnings("PMD.AvoidDuplicateLiterals")
+class JpMediaFileComparatorTest {
```
tesshucom, line 35: Asymmetry test cases will be added and coverage will be 100%. However in this time, Transitive test case is not added. It's a relatively easy fixes, and it's too disproportionate to the effort of creating a complecated test data.

Issue #1925, opened 2022-11-27
Redesigne scanning process


tesshucom:
Prerequisites: #1922, #1941, #1937, #1955, #1967, #1978, #1984, #2007

# Needs to be rewritten because it is obsolete

### Overview

A redesign of the scan is done. It will be a large-scale redesign, including related functions, instead of a partial refurbishment like before. Therefore, design changes are divided into several major versions.

Milestones | Main theme | Scheme Virsion Increment
-- | --  | --
v111.6.0 | Fix scanning workflow, latent bugs. | 
v112.0.0 | Scan status, Scan log viewer, Improve MusicFolder | required
v112.1.0 | Scan parallelization. Add video parser. | 
v113.0.0 | Podcast improvements | maybe needed

Version-No is provisional. Major version update will occur when fields in DB table changes.

<details>
<summary>The purpose
</summary>

 - There is a lot of waste in existing scan implementations. Just a correct rewrite will improve performance. Redesign make it easier to reduce waste, measurements, and add new features.
 - It will be designed to be very robust to data integrity. We aim for a server suitable for long-term operation that avoids data destruction.
 - Speedup isn't the primary goal, but turning on parallelism can improve things dramatically. on the other hand, parallelization feature will be provided as an additional option, not as a default.
   - Jpsonic also considers resource-constrained platforms such as Raspberry Pi, or NAS with embedded Java to be important targets. In these environments, it may be assumed that the scan parallelization option will not be used due to the user's operational policy. Taking full advantage of your machine's resources is sometimes a mistake.
   - Jpsonic's minimum requirements are the same as Subsonic's or less. (There are differences between 32-bit and 64-bit, so you should roughly double the memory estimate). But, even if you run it with the minimum requirements, it will still perform better than a traditional server.
   - Many of the bottlenecks are the performance of IO and network equipment. There may be some restrictions on using parallelism effectively. We aim to provide reasonable setting items and statistical information so that users can easily select the best effort according to users environment.
</details>

### Enumeration of expected work items

With so many fixes and verifications involved in scanning, where to draw the line is a thorny issue. This is a long term effort and may change from time to time.

<details>
<summary>Fixes coming in v111.6.0</summary>

 - The scanning process will be fragmented. And we will move to a design that focuses on delta updates.
   - Individual optimization and measurement will become easier. Although it seems complicated at first glance, the number of SQL issues at the time of new registration will be less than the legacy server.
[...]

PR #2041, opened 2023-02-19
Subdivide the scanning loop process

tesshucom:
Prerequisites: #1925

#### Overview

The loop processing that was done in [MediaScannerService#scanFile](https://github.com/airsonic/airsonic/blob/5ccca059d5cfe3dd19734c27861b434ea21b43d8/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java#L222) will be subdivided. Also includes some improvements.
Fragmentation will make it easier to improve individual logic and make it easier to locate when an exception occurs. 

<details>
<summary>Goals</summary>

 - The processes previously done in loops will be split into multiple smaller loop methods for each processing block. It is also changed to the correct processing order. Required for correct sort order and sort order related secondary processing. Legacy implementations generally parsed data in order of file operations, starting at the top node of the tree. Jpsonic is almost the opposite, parsing is performed from the leaves of the tree.
 - Elimination of the CreateOrUpdate method in Dao. This suggests that the conditions for creating and updating records are ambiguous, and become a barrier to speed improvement. They will be split into create and update and the conditions for calling them will be defined in the Service. This also relates to the Lucene index registration specification.
 - Remove unnecessary Lucene index entry registrations. In theory, if CUD successfully against the database, then should be able to do the same CUD against Lucene. 

</details>
<details>
<summary>Non-Goals</summary>

 - The main purpose is not to improve speed. This commit is intended to improve process consistency or design improvements to drive future improvements. 
 - Handling of sort tags is unchanged in this commit
 - The data flow is designed with scan cancellation in mind, but the cancel button has not been added yet
 - Some changes are expected for index display (web page) and cover art processing implemented at the same time as scanning. not changed yet in this commit. 

</details>
<details>
<summary>Notable behavior changes</summary>

 - In addition to the main storage of the database, the sonic server has an index storage for the search engine. Search index data exists in the data directory, with the directory name `**index**`. (In the very old Subsonic, the name is `lucene**`.)
 - A search index is a subrepository. If you stop the server and remove its directory, you will not be able to use the search or random list generation functionality. However, you can restore the search index by performing a scan.
 - If you deleted the index data, the scan was able to bring it back because it registers all the indexes during the scan.. After this commit, Jpsonic will do a delta update, reducing unnecessary writes to the repository. 
 - Jpsonic can do a full scan by **enabling the "ignore timestamps" option**. Even if there is no index data, all index data can be recovered if a full scan is performed. 
 - Traditionally, index data was used for searches and random lists. In Jpsonic, multiple-genres are realized by using index data, apart from Genre on the conventional database. Therefore, the designation of the genre is also affected.

</details>
<details>
<summary>Motivation</summary>

 - Eliminate OS-dependent operation specifications. It has to do with **two problems that many Subsonic users know**. In order to solve the following problems, it was necessary to divide the processing into blocks and reverse the processing order. 
   - Improve album parsing consistency. Conventional sonic servers parse files in order of appearance from the root of the file node. Simply put, in order of artist, album, and song. Parsing albums can be problematic. The old implementation transcribed the tag of the first child file encountered when parsing the album. This may or may not work, and may vary depending on the OS. (you won't notice this if the child files in the album have almost the same tags) Jpsonic refers to the track number and path in that order, and refers to the first fetched value as the basic information of the album. 
   - Change how album change detection works. "I rewrote the file, but the album data is doesn't change!" problem is relatively well-known. In that case, there is also a well-known bad solution that updating the date of the directory with the touch command will recognize the change. The cause of this is related to how the OS manages directory dates, and even the same OS may have different specifications depending on the version. (Performance-oriented OSes tend not to update the parent directory's date when a file changes. You may be able to change the default behavior through OS options.) This commit changes the parsing order so that it runs from the leaf instead of from the root node of the file. More emphasis is placed on the modification date of the child files rather than the date of the album directory which is OS dependent. Therefore, it is expected that the need to use the touch command will be eliminated.
 - Stats and genres no longer count increments while scanning. They are subprocesses rather than the main flow of scanning.
   - This commit will further tighten the unnecessary Genre removal logic.
   - Statistics management has already started to improve. #1954 When considering parallelization in the future, a design that accesses many optional shared variables concurrently is not a good idea.
   - In particular, Jpsonic realizes multi-genre, so the method of updating the genre master is slightly different. A method that aggregates the registered genre records at the end of the process is suitable.
 - **Each processing block has very different bottleneck characteristics**.
   - Each processing block requires a different approach and measurement to speed things up.
     - For example, reading a directory structure quickly may be related to the maximum number of file entries available in the OS. Especially for NAS.
     - Just because you have 32 CPUs doesn't mean you can analyze 32 multi-gigabyte videos at the same time.
     - On the other hand, there may be processing blocks that cannot be parallelized.
   - The division of processing blocks is very important for future design implovements. 

</details>
<details>
<summary>Description</summary>

The scanning workflow has changed, but the database design has not. However, the internal data specification is slightly different from Subsonic.

 - Subsonic's data update criteria is based on a comparison of file modification date and scan execution time. Jpsonic extends these and defines the literals FAR_PAST and FAR_FUTURE for dates.
   - FAR_PAST ; This is the so-called initial value. When this value and the scan execution date are compared, the scan execution date is always newer. Therefore, it indicates that the record for which this date is specified is always "subject to scanning". Can be used to force scanning of a particular album in addition to the initial value. Also changing the date of all records to this value will run a full scan.
   - FAR_FUTURE ; A flag value indicating that some future parsing is planned. For example, when parsing a File Structure, SONG is stamped with the scan execution time in the last_scanned field at the time of initial parsing. On the other hand, MOVIE is stamped with FAR_FUTURE. MOVIE is processed in a subsequent phase different from SONG, and will be rewritten from FAR_FUTURE to the scan execution time at parsing.

These additional specifications realize block division of scan processing and control of the processing order. Also, it has a high affinity with the cancellation specification. If the processing is interrupted at the appropriate place, the date flag will remain intact. The parsed or unparsed state of each file entry should be carried over to the next scan.

</details>


tesshucom (commented):
The fix has been completed. A simple operation check will be performed and if there is no problem, it will be rebase merged.

tesshucom:
Rebased due to being blocked by false positives for CVE-2022-45688.

tesshucom:
All changes rebased.

> Implement the processing equivalent to the following branch.
> https://github.com/airsonic/airsonic/blob/5ccca059d5cfe3dd19734c27861b434ea21b43d8/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java#L246-L250

Reconsider this.

tesshucom:
> > https://github.com/airsonic/airsonic/blob/5ccca059d5cfe3dd19734c27861b434ea21b43d8/airsonic-main/src/main/java/org/airsonic/player/service/MediaScannerService.java#L246-L250

≒ If scanning a podcast directory, index to ID3

The branching process above is a bit of bad logic. It will not be ported as an equivalent process.

 - Even with indexing, the registration data is never used, as most of the time the id3-index is a subquery of the music directory. ( = Podcast search is basically not possible)  Originally over-processed.
 - If we want to search for podcasts, there is a tacit knowledge that the podcast directory should be registered with the music directory. (I don't think such a specification is elegant, but it is the use case of the relatively accepted uses among Subsonic users.) 

≒ In other words, the condition for indexing Podcasr to ID3 can be replaced with the condition for whether a music directory with the same path as the Pocast directory is registered on the settings page. This commit will have an implementation with such condition.

 - This process is added at the beginning of ScannerProcedureService#**refreshAlbumID3**
 - This process is added at the beginning of ScannerProcedureService#**refreshArtistID3**

___

Note that in the case of the above design, there is no need for the following method to be split into two.

 - ScannerProcedureService#parseFileStructure
 - ScannerProcedureService#parsePodcast

This is a remnant of the legacy code that had independent loop operations. (Because legacy code uses an extra method parameter for Podcasts.)

In short, there is some code that could be simplified, but no further refactoring is done here.  That will take you nowhere... Podcast related fixes should be focused on [v111.8.0](https://github.com/tesshucom/jpsonic/milestone/63).

Issue #2008, opened 2023-01-26
Add new page to view detailed scan logs


tesshucom:
Prerequisites: #1978, 

Create a dedicated page for viewing scan logs.

 - Global log specification will be changed in v111.6.0.
   - #1978
   - Jpsonic has been improved not to crush the bubbling of exception handling. Therefore, it is assumed that the log level control by the external option works as expected to some extent. (I think the reason why the legacy log is redundant is that there is no certainty in the notification when a problem occurs, so it is stamped with clues before and after...)
   - With Sonic servers, users traditionally rely on logs to keep track of scan progress. In Jpsonic, if the log is used at the INFO level, the contents corresponding to the based on log previous spec will be output. Apart from that, it has been improved to optionally record scan logs in the database.

<details>
<summary>Example</summary>

Well, memory usage during scanning is probably less than similar servers..

```
select
	sl.type as scan_log_type,
	sl.start_date,
	se.type as scan_event_type,
	executed,
	max_memory / 1048576 as max_mb,
	total_memory / 1048576 as total_mb,
	-- free_memory,
	(total_memory - free_memory) / 1048576 as used_mb,
	comment	
from
	scan_log sl
left join 
	scan_event se
on
	sl.start_date = se.start_date
order by
	sl.start_date, se.executed
```

![image](https://user-images.githubusercontent.com/27724847/214769241-27ddb021-41d1-4f58-a162-8b99ef8e27ff.png)

</details>

Additional options allow memory metering. It is rare that such a function is added in logging. There are some special reasons:

 - This logging itself will change in v111.6.0. Parallelization of scanning and addition of a new video parser are planned for v111.7.0. Of course the defaults work as before. (It works stably enough on NAS and Raspberry Pi). Memory tuning will be essential when parallelizing or switching parsers. Measured values and guidelines for setting values should be presented to the user as simply as possible. (Perhaps there will be a wiki description related to this feature.)
 - Most new media servers these days run the risk of memory failures and crashes. Jpsonic has no such concerns so far, but it can be used as one to catch some symptoms. Of course, a good scenario would be to be able to prove that nothing happened.

So we need this web page from v111.7.0.

Issue #1373, opened 2022-03-02
Parallelization of processing (I)


tesshucom:
Prerequisites: #1925.

In v111.x.x, processing is parallelized step by step. 

Eventually, speedup will be achieved, but if anything, there is a big refactoring aspect of replacing each processing block with appropriate processing.

<details>
<summary>Notes remaining in #833</summary>

There had been some points to be noted when reviewing existing threads.

1. Scalability : 
1. Subdivision of scanning process : 
1. Inappropriate exception handling
1. Other aggressive parallelization : the system statistics page, `@Async`

Of these, 1-3 will be improved with v111.x.x. 

As for 4, speeding up the web page is less important for finding and listening to music, so we can postpone it.

____

When developing a new system, you should start with architect design. It's a refactoring of an existing system in production, so to a large extent the reverse procedure is taken. The following approach will be taken.

#### Scalability : 

Safety is more important than speed when parallelizing existing systems. A mechanism that allows the user to adjust Scalability is essential.

Most of the current scan bottlenecks are not the speed of the database but IO control costs. Parallelization can alleviate some of this, but in most cases threading above the number of cores is counterproductive. Therefore, apart from the existing pooling, a new pooling that can adjust the size according to the number of cores will be newly installed, and a function will be added so that the user can control it.

Most modern CPUs support hyper-threading, so Jpsonic will provide logical threads / 2 as a recommended value(That is, maybe the number of cores). 

 - The user must optionally be able to assign a value lower than the recommended value.
 - User can assign a value higher than the recommended value. However, this is counterproductive in most cases and should be warned(This is only needed if the core does not support hyper-threading).

The JVM does not handle these things automatically. Systems that do not care about this mechanism will easily experience CPU highload issues. Parallelization should not be implemented as long as such users lead to uncontrollable problems.

#### Subdivision of scanning process / Inappropriate exception handling

It is difficult to fix all of these at once. This is because the range is too wide. Therefore, the theme will be set for each version, and improvements, tests, and effect measurements will be carried out for each block.

In v111.2.0, the brush-up around the parser, which has already been partially started, will be done. In addition, parsing is parallelized so that parts of the entire scan can be executed in parallel. The design of the newly created thread pool, queuing, and task is provisional, and if the verification and modification of other processing blocks are completed later, the task design may change cumulatively.

</details>

#### Fixes

Related bugs need to be fixed first.

- [x] #1376
- [x] #1392
- [x] #1405

v111.2.0
_____
[...]

PR #2079, opened 2023-03-12
Add scan cancel button

tesshucom:
Prerequisites: #1925, #1373, #1978

#### Overview

When forcing the user to do a lengthy process with no expected end time, it is common to indicate an estimated time or make it cancelable. The process of estimating the time is costly and is putting the cart before the horse. A cancel button should be provided.

#### Goals

Allow users to cancel from the same web page after starting a scan. And displays warnings if the previous scan did not complete successfully.

#### Non-Goals

 - Cancellation is possible, but since it is a process that forcibly interrupts the scan, it is necessary for the user to be aware that it will end in an incomplete state.
   - It is not a function that saves the intermediate state of the flow and resumes from there
   - It's not a feature to rollback parsing
   - For example, if the parse is half done, parsed files will not be parsed on next scan. All it provides is a premature stop and retry. It is assumed that the consistent state will be restored by executing the scan to the end again.
 - Detailed logs will not be provided on the web page in the next release version. The next next version will provide a more detailed tracking page.
   - #2008

#### Motivation

The use case is primarily for emergency shutdowns. For example, when the power supply of NAS switches to UPS during a power failure and a safe shutdown is attempted. Subsonic and Airsonic don't have such a case and we don't know what will happen.

Jpsonic already includes a fix that stops the scanning process on shutdown, making it less likely to destroy data integrity. This pull request is a fix to extend that functionality and allow users to cancel at any time from the web page.


tesshucom:
Hmm, maybe status judgments are wrong? Operation confirmation required.

tesshucom:
Have understood. It's not about the cancellation per se.

There seems to be a missing case in the current judgment of skippability. If a directory is deleted and recreated, etc., simply referencing whether the previous result was successful will result in false positives. 

 -> This issue requires another additional implementation. This pull request completes itself.

PR #2091, opened 2023-03-19
v111.6.0 RELEASE

tesshucom:
Since there are many changes, Breaking Changes, and Materials compiled based on Web page viewpoints will be presented on the release page.

 - [x] Create CHANGELOG.
 - [x] Summary of some breaking changes.
   - All of them are important if you rank them according to their so-called importance, but the following need special attention. Because they are related to Java startup commands. It will be a problem before starting the application. The following announcement will be added to the release page.
     - **Default log level changed**
       - > Change default log level from INFO to WARN. The number of logs output to the console and files is drastically reduced.
       - > You can also revert to the previous log level. For standalone, the log level can be changed with the startup argument.
`-Dlogging.level.com.tesshu.jpsonic=INFO`
For Docker, configuration items will be provided in production.yml.
       - Click here for details : #1978 
     - **Major version upgrade of search engine**
       - > Without the -Xmx option, a bit more buffer may be reserved than 8.x. -Xmx512m, etc. By setting it appropriately, it should be possible to use it with almost the same amount of resources as before.
       - In the case of 100,000 songs or less, it is assumed that operation is possible at 512m. (In practice, it will run below that.) Of course you can allocate more memory, but I don't know if it will lead to good results. Starting with v111.6.0, Jpsonic adds an option to track memory usage during scanning. This is currently a developer feature, but a web page will be added to view this in the next version.
       - Click here for details : #2061 
 - [x] Materials compiled based on Web page
   - #2092


tesshucom:
Codacy Static Code Analysis complains about the Diff/Deltas process, but it is ignored.

tesshucom:
This pull request may be discarded and recreated anew.

I have some concerns, so I'm checking the operation. Since it will be a version upgrade with a considerable impact internally, it will be released carefully.

PR #2117, opened 2023-04-01
Migrate from dbcp2 to HikariCP

tesshucom:
Will be migrated from dbcp2 to HikariCP.

 - Bump HikariCP from 4.0.3 to 5.0.1
 - Migrate from dbcp2 to HikariCP
 - Remove commons-dbcp2

Values previously used for pooling will be inherited for safe migration.

<details>
<summary>Previously Adopted Value</summary>

Main field values of dbcp2. Note that default values are used except for the APP value, but Jpsonic relies on those default values. So the same values will be mapped to timeouts, Idle sizes, etc in this version.  If future adjustments are necessary, they will be considered at that time.

In addition, note that the DefaultQueryTimeout is set to Null. In this case the database default value will be adopted. HSQLDB uses a maximum timeout of 60 seconds if timeout has been specified as zero. So by specifying 60 seconds, we can avoid a lot of confusion.

Fields | Values
-- | --
LoginTimeout | UnsupportedOperationException
DriverClassName | **APP value**
Url | **APP value**
Username | **APP value**
Password | **APP value**
AbandonedUsageTracking | FALSE
AutoCommitOnReturn | TRUE
CacheState | TRUE
ConnectionInitSqls | EMPTY
DefaultAutoCommit | NULL
DefaultCatalog | NULL
DefaultQueryTimeout | NULL
DefaultReadOnly | NULL
DefaultSchema | NULL
DefaultTransactionIsolation | -1
DisconnectionSqlCodes | EMPTY
EvictionPolicyClassName | org.apache.commons.pool2.impl.DefaultEvictionPolicy
FastFailValidation | FALSE
InitialSize | 0
JmxName | NULL
Lifo | TRUE
LogAbandoned | FALSE
LogExpiredConnections | TRUE
MaxConnLifetimeMillis | -1
MaxIdle | 8
MaxOpenPreparedStatements | -1
MaxTotal | 8
MaxWaitMillis | -1
MinEvictableIdleTimeMillis | 1800000
MinIdle | 0
NumActive | 0
NumIdle | 0
NumTestsPerEvictionRun | 3
RemoveAbandonedOnBorrow | FALSE
RemoveAbandonedOnMaintenance | FALSE
RemoveAbandonedTimeout | 300
RollbackOnReturn | TRUE
TestWhileIdle | FALSE
TimeBetweenEvictionRunsMillis | -1
ValidationQuery | NULL
ValidationQueryTimeout | -1
</details>

In this migration, datasource instantiation does not rely on Spring's builder. There is a possibility that it will be tuned when parallel scan execution is introduced in the future, but at that time we will reconsider whether to move to Spring's Builder. 

 - It is smarter to provide default values that rarely need to be changed.
 - JNDI allows the user to configure it however they like.

### DIFF
--- a/jpsonic-main/pom.xml
+++ b/jpsonic-main/pom.xml
@@ -80,6 +80,17 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-jdbc</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.zaxxer</groupId>
+                    <artifactId>HikariCP</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>com.zaxxer</groupId>
+            <artifactId>HikariCP</artifactId>
+            <version>5.0.1</version><!--$NO-MVN-MAN-VER$-->
         </dependency>
         <dependency>
             <groupId>org.springframework.boot</groupId>
--- a/jpsonic-main/pom.xml
+++ b/jpsonic-main/pom.xml
@@ -194,19 +194,6 @@
             <scope>runtime</scope>
         </dependency>
 
-        <!-- connection pool manager, fixes #860 (external database performance) -->  
-        <dependency>
-            <groupId>org.apache.commons</groupId>
-            <artifactId>commons-dbcp2</artifactId>
-            <!-- this wants to pull in commons-logging v1.2 which conflicts with other dependencies -->
-            <exclusions>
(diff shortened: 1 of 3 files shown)


tesshucom:
[2023-04-01 09:50:45] [autobuild] Caused by: org.apache.maven.wagon.TransferFailedException: transfer failed for https://repo.maven.apache.org/maven2/org/directwebremoting/dwr/3.0.0-rc3-RELEASE/dwr-3.0.0-rc3-RELEASE.pom

PR #2184, opened 2023-05-11
Provide automatic updates for Docker images

tesshucom:
Prerequisites: #2175

Targeting a Synology DS220+ or general Linux, docker-compose.yml from watchtower will be provided for automatic Docker updates. If Docker is started as per #2175, you can schedule automatic updates with one command.


tesshucom:
@dependabot rebase

PR #2270, opened 2023-06-27
Update Libs and Podcast input check improvements

tesshucom:
This patch does not contain critical security lib updates.

 - Includes false positive suppression for CVE-2023-35116.
 - Podcast episodes will be modified to check the extension before downloading.
   - #2265


tesshucom:
It will be recreated due to lack of consideration of one point

Issue #2305, opened 2023-08-04
End of Java 11 support


tesshucom:
## Overview

Jpsonic's Java11Support will end when v112.2.0 is released.

### Detail

Now is a compelling time to end support for Java 11.

 - The end of Active Support for Java11 in Sep 2023
 - Required to prepare for [Spring Boot 3.0](https://github.com/tesshucom/jpsonic/issues/839)
   - Open source support for Spring 2.x will end in Nov 2023
 - Required for [DWR upgrade](https://github.com/tesshucom/jpsonic/issues/1630)
 - [JDK21 GA release in September 2023](https://openjdk.org/projects/jdk/21/).
   - For Oracle, this version is the next LTS. In other words, Java 11 has come to an end, and it is time to search for the next mainstream.
 - Jpsonic's Docker is steadily improving. It will continue to improve. 
   - In other words, it can be used even on OSs that do not provide the latest Java.

![image](https://github.com/tesshucom/jpsonic/assets/27724847/c21cb1a2-0e30-4dda-b6d0-b1ba07468a85)

PR #2327, opened 2023-08-22
Remove Java11 from CI

tesshucom:
Prerequisites: #2305

### Overview

The end of Active Support for Java11 in Sep 2023. Along with this, Jpsonic's Java11 Support will also end, so it will be removed from CI.

### Note

 - Some of our CI will continue to use Java 11, as the tools that submit coverage to Codacy only support Java 11. It doesn't affect the build of Jpsonic .
 - This pull request also serves as a working confirmation of Sonatype-lift. Sonatype-lift will be removed, especially if Facebook Infer does not work. (In any case, Sonatype-lift has been announced to end. It will be replaced by others in the near future. There was also a notice. `A new offering launching in late-2023: Sonatype Developer`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,16 +55,16 @@ jobs:
       uses: actions/setup-java@v3
       with:
         distribution: 'temurin'
-        java-version: 11
-    - name: Autobuild
-      uses: github/codeql-action/autobuild@v2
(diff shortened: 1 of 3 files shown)


tesshucom:
Sonatype-lift was already stopped so there was no problem, but CodeQL doesn't seem to work.

Issue #2346, opened 2023-09-04
Agenda for v113


tesshucom:
### Overview

The theme of the next version is to speed up index generation. It may seem like scanning is suddenly off-topic, but it's not. Index generation can be streamlined by including it in the scanning procedure. Therefore, it is essentially a change in scanning.

### Problem description

Index generation is too slow.

<details>
<summary>How slow is it?</summary>

[TestDataCreator.zip](https://github.com/tesshucom/jpsonic/files/12832002/TestDataCreator.zip)

Code to create test data for Music Index. In Git, it may exist in a sandbox branch, but it is never merged. 

___

Create data for Artists 8K, Songs 400K. Adjust index count to 70. The combination of Japanese and English would probably be close to that number. Or an index with a mix of three or so languages might be this size.

![image](https://github.com/tesshucom/jpsonic/assets/27724847/5fd59905-8f1e-4a40-b3b2-6356dd05bc5a)

This kind of index is certainly not common in music software. However, Japanese people don't find it so unnatural. Touch panels similar to this are used in banks and public institutions. Because, the number of keystrokes is reduced to allow for use by elderly people who are not familiar with PCs, or by people with handicaps.

If the artist is 8K, the index generation time will be longer. Of course. When I measured it, it was as follows (this is not the top page display time. It is the index generation time, in Service class)

6,493 ms
2,126 ms
1,977 ms
1,843 ms
1,625 ms
1,522 ms
1,548 ms
1,555 ms

Very slow if not optimized. That's why the first few times are slow.

This is the time on a DS220+, but it will probably be slower on a slower PC. There were reports that it took several minutes to log in with Airsonic, and that's probably not an exaggeration. The environment specs are low, it is not tuned, etc. Also, GC may be occurring due to low memory or other reasons. The time required will definitely increase at an accelerating rate. 

No matter how fast you do it, it takes 1.5 seconds, so no matter how you think about it, it's not something that will be placed on the top page. Improvements in this regard require a two-pronged approach.

 - Speeding up index generation itself : This is because the Subsonic API requires a Full size Index.
   - This is mainly handled in v113.0.0
 - Web page improvements : This will be addressed in the derivation of #2349
   - Although the menu is planned to be reorganized, individual Web pages will also need to be reimplemented.

</details>

 - 1.5 seconds - 6 seconds may seem like a tolerable value, but that's not the case.
   - On a web page, various processes are concentrated during login immediately after startup. Those 6 seconds are fatal.
[...]

PR #2422, opened 2023-10-22
Fix for generating MusixIndex when scanning

tesshucom:
### Overview

Jpsonic reimagines the lifecycle of music indexes to be more or less scan-based rather than real-time. Part of the music index creation is included in the scan, allowing faster generation of the music index at the time of your request.

### Known issues

With Subsonic and Airsonic, music indexes are generated in real time on request. Therefore, when a request occurs, a series of processes are executed in a chain: obtaining the artist, indexing based on the name, and generating an index object that holds a large number of entries. It is inherently a fairly expensive process, and not a problem for small libraries, but it becomes a problem for large music libraries. 

#### Examples that are likely to cause problems

 - After scanning a large library, the web page freezes when restarted and logged in
 - Below the DLNA communication speed target

### Approach

Jpsonic "pays up front at the time of scanning" processing costs that are frequently used even though speed is required, such as sorting and index generation. Although the scanning process is complex, its accuracy is more reliable.

Unlike Subsonic and Airsonic, Order and Musix-Index can be used with SQL. This suggests that front-end implementation will be easier. Of course, the processing speed will be **faster**, and it will be easier to create **views that were previously impossible**.

### Goal

 - Change DB schema. In the media_file and artist tables, a field called music_index will be added
 - Fix the index to be registered correctly when scanning.
 - Scan speed will be adjusted so that it is no longer slower than before.

### Non-Goal

Given the large size of the update, this pull request only includes fixes related to the indexing process. Fixes for UPnP, Rest, and Web pages will be done separately.

### Details

In addition to indexing, there are two other relatively important changes. MusicIndexService related fixes and speed improvements.

 - Create an interface for MusicIndexService. 
   - MusicIndexService had quite a lot of methods. However, by design, only the three that are for reference only should be published.
   - Most of the logic should be hidden in the scanner package.
 - Improvements to the differential update method for Artist's sort tag
   - There are multiple artist-related sort tag fields scattered in media_file. Artist, album artist, composer. Previously, a separate query was issued to correctly modify these values. If Music-Index is added to this, the number of issued queries will become enormous. Therefore, it will be modified to merge the difference contents in advance and update the data rows with one query.

MusicIndexService's Interface method is now the simplest. This pull request does not include any other fixes, but there is a high possibility that it will change in the future.

 - Legacy's domain class MusicIndex includes Comparator. This is likely overkill for an interface. Probably these will be removed.
 - It is likely that methods will be added to MusicIndexService. Instead of a full index Map of MusicIndex and Artist, a method to obtain a lighter Map of MusicIndex and number of records, etc.

### Impact on scan speed

The bottom line is that the speed improvements will outweigh the increased costs, so scanning large libraries will be faster than before.

🙄🙄🙄　

___

Factors causing speed reduction

 - One field each will be added to media_file and artist_table. They are also given an index (of DB Scheme).
 - Parsing is done to determine the index string during the scan

These will mainly affect three scan procedures.

 - PARSE_FILE_STRUCTURE 
   - On the settings page, users can optionally skip the UPDATE_SORT_OF_ARTIST step. In that case, indexing will occur during this step. (Not done by default)
   - A large number of entries are made to the media_file table (effects due to increase in fields and addition of index)
 - UPDATE_SORT_OF_ARTIST 
   - By default, the music-index is determined in this step. This is because tag correction processing is performed after all file analysis is completed. Creating a music-index requires the name and its pronunciation, so if the design is correct, the Music-Index cannot be determined without this step.
 - REFRESH_ARTIST_ID3 
   - Subsonic and Airsonic have problems with ID3 data registration. Noisy-fake data may be created due to the file structure. After Jpsonic completes data registration and fix of all Song files, it fetches those tags and merges them to the Artist and Album tables. Therefore, ID3 Artist's Music-Index will be generated from the Song's Tag in this process, not by reusing the File Structure. 
   - Effects due to increase in fields and addition of index, and parsing.

The assumed data is explained in Agenda.

 - #2346

> Create data for Artists 8K, Songs 400K. Adjust index count to 70. The combination of Japanese and English would probably be close to that number. Or an index with a mix of three or so languages might be this size.

Furthermore, to eliminate the influence of HDD seek time, dummy data of playing time = 0 will be used. (It's probably close to the theoretical value of a Full SSD.) No sort tags are registered in the file. This is the pattern that takes the most time to analyze.

112.2.8 | 1 | 2 | 3
-- | -- | -- | --
All | 0:23:49.448 | 0:23:31.212 | 0:23:30.638
PARSE_FILE_STRUCTURE | 0:10:01.323 | 0:10:12.800 | 0:10:11.213
UPDATE_SORT_OF_ARTIST | 0:05:31.063 | 0:05:27.152 | 0:05:27.574
REFRESH_ARTIST_ID3 | 0:00:26.489 | 0:00:26.777 | 0:00:26.978

This Pull request | 1 | 2 | 3
-- | -- | -- | --
Duration All | 0:23:05.734 | 0:23:12.914 | 0:22:58.132
PARSE_FILE_STRUCTURE | 0:10:05.935 | 0:10:21.528 | 0:10:12.039
UPDATE_SORT_OF_ARTIST | 0:04:56.601 | 0:04:53.224 | 0:04:50.889
REFRESH_ARTIST_ID3 | 0:00:28.619 | 0:00:27.826 | 0:00:28.300

🙄🙄🙄 

Normally, the scan time will be slightly longer. However, UPDATE_SORT_OF_ARTIST has become a little faster, so the overall time has become shorter.

### Impact on users

 - A full scan will be required for next release versions.
 - Previously, if you changed the string used for the index on the settings page, it would be reflected by pressing F5. After this pull request, changing the index string will require a full scan.
   - It's probably not something that changes often, and the settings appear to have a very long lifecycle. It is unlikely to be considered a de facto functional restriction.
 - With the default settings on the first scan, the music-index is not usable until the UPDATE_SORT_OF_ARTIST step is reached. This makes sense when you think about it logically.
   - Subsonic certainly seems to have excellent responsiveness, but there is a fatal problem lurking in the operational design of its View.
     - Depending on the settings, if the data does not exist, it will search for data on the file system and create the data. And it will run on the unbounded http thread rather than the scanning thread.
   - Jpsonic will not display the data until it is available. That's all.

The third impact has no effect at the time of merging this pull request. There will be an impact when speeding up the View in subsequent pull requests.


jpsonic-main/src/main/java/com/tesshu/jpsonic/domain/JapaneseReadingUtils.java
```diff
@@ -382,13 +381,25 @@ private ReadingResult toReading(Token token) {
...
+        }
+        String result = s;
+        for (String article : settingsService.getIgnoredArticlesAsArray()) {
+            if (StringUtils.startsWithIgnoreCase(s, article + SPACE)) {
+                result = result.substring(article.length() + 1) + ", " + article;
+            }
+        }
+        return result;
+    }
+
```
tesshucom, line 402: Consideration of rules regarding article removal. Rotation is sometimes seen as a method of processing articles that gets in the way when sorting in Latin. In that method, all entries are converted to "The Beatles -> Beatles, The" and then sorted. Personally, I don't think it's a very meaningful cost, and I think it would be better to simply delete it. (We cannot imagine any realistic case where this difference would cause a problem. Of course, if such cases are discovered in the future, they can be fixed at that time.)

Newly created sorts in Jpsonic are already unified to simple deletion. Rotation processing has been maintained to preserve Airsonic's implementation, but there is no need for rotation when generating Music-index. It looks like the string rotation process can now be removed.

tesshucom: I think that the existence of this kind of logic is probably a remnant of Pgrammer's classification done at a primitive stage, asking whether the article was deleted or whether the string was originally a string without an article. There is a high possibility that this is one of the bad habits rather than clean logic. (By the way, Japanese always uses Non-article noun, so maybe I'm just insensitive to this issue and it just looks like debug code.)

jpsonic-main/src/main/java/com/tesshu/jpsonic/dao/MediaFileDao.java
```diff
@@ -865,56 +870,96 @@ select count(id)
...
                     where id = ?
-                    """, candidate.getSort(), candidate.getId());
+                    """, cand.getSort(), cand.getTargetId());
         }
     }
 
+    public void updateArtistSort(List<ArtistSortCandidate> cands) {
+        String updates = "update media_file ";
+        StringBuilder cols = new StringBuilder("set ");
+        String cond = "where id = :id";
```
tesshucom, line 932: Change to '?'

jpsonic-main/src/main/java/com/tesshu/jpsonic/dao/MediaFileDao.java
```diff
@@ -865,56 +870,96 @@ select count(id)
...
+            }
+            case UNKNOWN -> {
+            }
+            default -> {
+            }
+            }
+        });
+        args.add(cands.get(0).getTargetId());
+
+        String query = updates + cols.toString().replaceFirst(",$", "") + cond;
```
tesshucom, line 959: Make whitespace replacement a little stricter(Maybe it doesn't work with Postgres)

jpsonic-main/src/main/java/com/tesshu/jpsonic/dao/MediaFileDao.java

jpsonic-main/src/main/java/com/tesshu/jpsonic/domain/JapaneseReadingUtils.java
tesshucom: Done.

Issue #1879, opened 2022-11-12
Faster index generation


tesshucom:
Webpage considerations. Related airsonic/airsonic#881,  #1734.
Prerequisites: #1941, #1937. #1986, #1925

### Overview

Speed ​​up index display. Index generation logic will be redesigned.

<details>
<summary>Known issues</summary>

The logic involved in indexing had been implemented as a submethod of the index method. Therefore, when a request occurs, a series of processes are executed in a chain: obtaining the artist, indexing based on the name, and generating an index object that holds a large number of entries. It is inherently a fairly expensive process, and not a problem for small libraries, but it becomes a problem for large music libraries. 

 - It will become a bottleneck when managing hundreds of thousands of songs or running on low-spec machines where media servers are often deployed.
 - There is also a problem with View.
   - Airsonic's Web Page has a unique design in which the index is displayed on the top page.
     - Generally speaking, hierarchical browsing functionality should be provided. Considering the use of screen readers, putting thousands of links on the top page seems a bit brutal.
     - ~~We will try to keep Airsonic-flavored browsing. It will be moved to deprecated features.~~
       - At this stage, this UI is very likely to be scrapped. It will be replaced by a relatively similar UI that is less risky in large libraries.
   - A huge for statement is included from index generation to display. Especially before hotspots are applied ... the first run after startup can be catastrophically slow.
     - This phenomenon is demonstrated in Web Page. 
     - UPnP is especially prone to problems. If DLNA Player is used, there is a high probability that it will not be able to meet the response speed requirements. (Jpsonic has a rather complicated cache implementation to avoid this. It's better not to have it.)

</details>

### Goal

 - Keep the index generation response for a library of 400,000 songs below 400ms

<details>
<summary>Approach</summary>

There will be a redesign that incorporates some of the logic required for index generation into the scan design.

The old IndexService defines all the processes of analyzing the data to be registered in the database and creating an index. It is efficient to break this down. 

#### Create, Update, Delete

 - Add index field to media_file, artist table
 - When registering or updating media_file and artist records, determine the value of the index field. In other words, indexing is done at scan time. 
 - The index can be fixed after the "reading" of the artist name is fixed. That is, after a series of sort tag modification processes. Therefore, it needs to be included in the initial parse, merge, copy, and completion processes. This cost is expected to be very small.

#### Read

 - By using the above index field together with Jpsonic's order field, it will be possible to generate an index with one SQL query.
[...]

PR #2436, opened 2023-11-02
UPnP design improvements

tesshucom:
Prerequisites: #1879 

### Overview

UPnP will be redesigned.

 - This will be a prerequisite for improvements will made in v113.0.0. Music index caching mechanism will be removed in v113.0.0
 - A later version is planned to allow the user to name display items and change the order. Web pages and UPnP menu items will be customizable
 - New items such as ID3 format Genre will be added.

The scanning fix is relatively nearing completion, so it is likely that web pages and UPnP will be improved concurrently in the future. Before we can do that, we'll need to clean up our UPnP implementation a bit.

### Details

Since a considerable amount of rewriting will occur, the details will be omitted. At least the following points should be corrected.

 - Correctly modify class design. Eliminate circular references
   - Jpsonic reused some of Airsonic's tricks, but also inherited some of the bad stuff (circular references). Dispatcher must not be exposed in UpnpContentProcessor. Must be RootProcessor limited logic.
     - [getDispatcher ... ](https://github.com/airsonic/airsonic/blob/5ccca059d5cfe3dd19734c27861b434ea21b43d8/airsonic-main/src/main/java/org/airsonic/player/service/upnp/UpnpContentProcessor.java#L150C23-L152)
 - Organize Processor IDs. String will be changed to Enum. The logic for generating UPnP display names will change.
   - UPnP display name is obtained from property resource. (English or Japanese exist) Many UPnP Processors use lazy loading, so this initialization mechanism is a bit complicated (though it works with the current implementation. If make a slight mistake, it won't work.). In addition, they are managed independently for each processor.
   - Rather than doing naming processing internally in Processors, it will be changed to a method that relies on ID and is set from the outside. The mechanism will be relatively simple. UPnP display name and order will be added in later versions so that users can change them. It would be better to improve the design now.
 - Add UpnpDIDLFactory
   - Among the processes that convert domain objects to UpnpDIDL, those that can be shared by multiple Processors will be modified to use Factory. Improved readability? No.
   - Tree nodes are typically represented hierarchically starting from menus and eventually reaching media files. This will not be programmed as is. The ID of the DIDL at the end of those tree nodes suggests chane processing by other Processors.
     - The design would not be acceptable if it were the server's responsibility to generate multiple complete strict trees. However, in typical UPnP apps, data searches are only performed in the forward direction from root to leaf. I have never seen an app that searches backwards. If the Client App will only explore Treenodes in the forward direction, the design can be simplified.

### Goal

The design will be improved so that functionality already provided is not affected.

### Non-Goal

Fixes related to the music index are not included in this commit.

### Note

Basically the feature remains the same. However, the specifications will change in the following points:

#### Fixed to not output empty albumArtURI

The previous implementation intentionally outputted an empty entry if no cover art was present. This will be removed from this commit.

```<upnp:albumArtURI/>```

This is not a defect. There was an app where the display changed depending on whether there was an empty entry or not, so this had been output as a workaround. (The displayed icon changes. Instead of an Artist icon without an image, to a folder etc. ) Now it seems that it is no longer confirmed so the extra output will be removed.

### CODE REVIEW

**Comment by tesshucom** [comment]
Well... commit length is too large for importance, so it will be squashed after review.

File: jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/UPnPContentProcessor.java [hunk 1]
```diff
@@ -0,0 +1,113 @@
...
+        return createBrowseResult(content, 1, 1);
+    }
+
+    Container createContainer(C containerEntity);
+
+    default void addItem(DIDLContent parentContainer, C containerEntity) {
+        parentContainer.addContainer(createContainer(containerEntity));
+    }
+
+    List<C> getItems(long offset, long maxLength);
```
tesshucom, line 76: Better change the name.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/AlbumByGenreUpnpProcessor.java
```diff
@@ -22,88 +22,68 @@
...
-                + item.getId());
+        container.setId(ProcId.FOLDER.getValue() + ProcId.CID_SEPA + item.getId());
         container.setTitle(item.getName());
         container.setChildCount(getChildSizeOf(item));
         if (mediaFileService.isRoot(item)) {
-            container.setParentID(getRootId());
+            container.setParentID(getProcId().getValue());
         } else {
             MediaFile parent = mediaFileService.getParentOf(item);
             if (parent != null) {
```
tesshucom, line 89: Add test.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/AlbumByGenreUpnpProcessor.java
```diff
@@ -147,26 +138,24 @@ public MediaFile getItemById(String id) {
 
     @Override
     public int getChildSizeOf(MediaFile item) {
-        return searchService.getAlbumsByGenres(item.getName(), 0, Integer.MAX_VALUE, util.getGuestMusicFolders())
-                .size();
+        return searchService.getAlbumsByGenres(item.getName(), 0, Integer.MAX_VALUE, util.getGuestFolders()).size();
```
tesshucom, line 141: Add count method to Dao.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/ArtistByFolderUpnpProcessor.java
```diff
@@ -101,35 +99,22 @@ protected static final String createArtistId(String id) {
...
+            container.setId(getProcId().getValue() + ProcId.CID_SEPA + item.getId());
+            container.setParentID(getProcId().getValue());
             container.setTitle(item.getName());
             container.setChildCount(item.getArtist().getAlbumCount());
             if (item.getArtist().getCoverArtPath() != null) {
-                container.setProperties(Arrays.asList(
-                        new ALBUM_ART_URI(getDispatcher().getArtistProcessor().createArtistArtURI(item.getArtist()))));
+                container.addProperty(factory.toArtistArt(item.getArtist()));
             }
             return container;
```
tesshucom, line 109: Check if it can be shared.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/ArtistUpnpProcessor.java
```diff
@@ -19,73 +19,65 @@
...
         MusicArtist container = new MusicArtist();
-        container.setId(getRootId() + UpnpProcessDispatcher.OBJECT_ID_SEPARATOR + artist.getId());
-        container.setParentID(getRootId());
+        container.setId(getProcId().getValue() + ProcId.CID_SEPA + artist.getId());
+        container.setParentID(getProcId().getValue());
         container.setTitle(artist.getName());
         container.setChildCount(artist.getAlbumCount());
         if (artist.getCoverArtPath() != null) {
-            container.setProperties(Arrays.asList(new ALBUM_ART_URI(createArtistArtURI(artist))));
+            container.addProperty(factory.toArtistArt(artist));
```
tesshucom, line 68: Check if it can be shared.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/ArtistByFolderUpnpProcessor.java
```diff
@@ -144,10 +129,9 @@ protected static final String createMusicFolderId(String id) {
     @Override
     public List<FolderArtistAlbumWrapper> getChildren(FolderArtistAlbumWrapper item, long first, long maxResults) {
         if (item.isArtist()) {
-            return getDispatcher()
-                    .getAlbumProcessor().getAlbumsForArtist(item.getName(), first, maxResults,
-                            util.isSortAlbumsByYear(item.getName()), util.getGuestMusicFolders())
-                    .stream().map(FolderContent::new).collect(toList());
+            return albumDao.getAlbumsForArtist(first, maxResults, item.getName(),
+                    util.isSortAlbumsByYear(item.getName()), util.getGuestFolders()).stream().map(FolderContent::new)
+                    .collect(toList());
```
tesshucom, line 134: No need to use Collectors anymore.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/DispatchingContentDirectory.java
```diff
@@ -197,99 +181,37 @@ public BrowseResult search(String containerId, String upnpSearchQuery, String fi
...
 
         if (Artist.class == criteria.getAssignableClass()) {
-            return getArtistProcessor().toBrowseResult(searchService.search(criteria));
+            return artistProcessor.toBrowseResult(searchService.search(criteria));
         } else if (Album.class == criteria.getAssignableClass()) {
-            return getAlbumProcessor().toBrowseResult(searchService.search(criteria));
+            return albumProcessor.toBrowseResult(searchService.search(criteria));
         } else if (MediaFile.class == criteria.getAssignableClass()) {
-            return getMediaFileProcessor().toBrowseResult(searchService.search(criteria));
+            return mediaFileProcessor.toBrowseResult(searchService.search(criteria));
```
tesshucom, line 188: Move toBrowseResult to Factory class.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/RootUpnpProcessor.java
```diff
@@ -21,134 +21,172 @@
...
-        root.setId(UpnpProcessDispatcher.CONTAINER_ID_ROOT);
+        root.setId(ProcId.ROOT.getValue());
         root.setParentID("-1");
-
         // MediaLibraryStatistics statistics = indexManager.getStatistics();
         // returning large storageUsed values doesn't play nicely with
         // some upnp clients
         // root.setStorageUsed(statistics == null ? 0 :
         // statistics.getTotalLengthInBytes());
         root.setStorageUsed(-1L);
```
tesshucom, line 78: Comments are not needed in code in most cases.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/SongByGenreUpnpProcessor.java
```diff
@@ -80,16 +75,16 @@ public Container createContainer(Genre item) {
 
     protected Container createContainer(Genre item, int index) {
         GenreContainer container = new GenreContainer();
-        container.setId(getRootId() + UpnpProcessDispatcher.OBJECT_ID_SEPARATOR + index);
-        container.setParentID(getRootId());
+        container.setId(ProcId.SONG_BY_GENRE.getValue() + ProcId.CID_SEPA + index);
+        container.setParentID(ProcId.SONG_BY_GENRE.getValue());
         container.setTitle(util.isGenreCountAvailable()
                 ? item.getName().concat(" ").concat(Integer.toString(item.getSongCount())) : item.getName());
         container.setChildCount(item.getSongCount());
```
tesshucom, line 82: Use getRootId()

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/UpnpDIDLFactory.java
```diff
@@ -0,0 +1,235 @@
...
+
+    private static final String SUB_DIR_EXT = "/ext/";
+
+    public UpnpDIDLFactory(SettingsService ss, JWTSecurityService jwt, MediaFileService ms, PlayerService playerService,
+            TranscodingService ts) {
+        settingsService = ss;
+        jwtSecurityService = jwt;
+        this.mediaFileService = ms;
+        this.playerService = playerService;
+        transcodingService = ts;
```
tesshucom, line 79: Avoid using ShortName as much as possible in constructor.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/UPnPContentProcessor.java
```diff
@@ -0,0 +1,113 @@
...
+ * An interface for representing a typical UPnP node tree with domain object.
+ *
+ * @param <C>
+ *            Container entity. Most are domain objects used for hierarchical expression such as Artist, Album,
+ *            MefiaFile, etc. Or, there may be cases where multiple domain objects are composited and used as the
+ *            Wrapper.
+ * @param <I>
+ *            Item entity. Represents a Chidren provided by a Container. At the end of Leaf, it becomes a media file
+ *            that represents music or video. For browsing that provides many hierarchies, this entity may be a
+ *            container.
```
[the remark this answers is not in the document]
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/UPnPContentProcessor.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/AlbumByGenreUpnpProcessor.java
tesshucom: Ah.... This was implemented in Lucene. And it is assumed that it is cached and the correct offset count is used. We can leave it as is unless it causes a fatal problem.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/ArtistByFolderUpnpProcessor.java
tesshucom: I'm certainly curious, but there were 73 locations, so I'll probably be able to rewrite it after closing this pull request.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/RootUpnpProcessor.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/UpnpDIDLFactory.java
tesshucom: Done

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/SongByGenreUpnpProcessor.java
tesshucom: Done,

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/DispatchingContentDirectory.java
tesshucom: Hmm, at first glance it seems like this can be made common, but it might be better not to do it yet. The class that currently returns BrowseResult is clearly a subclass of  UPnPContentProcessor. DIDLContent, which is an intermediate object, may be transferred to Factory, but if BrowseResult is generated by Factory, the responsibilities between classes will be ambiguous.

tesshucom: This is not fixed now.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/ArtistByFolderUpnpProcessor.java
tesshucom: This pull request will only target upnp package and below.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/ArtistByFolderUpnpProcessor.java
tesshucom: Done. However, this problem will not be resolved until the music index is fixed. It will be completely standardized later.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/AlbumByGenreUpnpProcessor.java
tesshucom: Yeah, I think this will be done when DIDL is standardized when MusicIndex is redesigned. It will be checked at that time.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/IndexId3UpnpProcessor.java
```diff
@@ -138,10 +136,11 @@ public Container createContainer(Id3Wrapper item) {
             MusicAlbum container = new MusicAlbum();
             Album album = new Album();
             album.setId(id);
-            container.setAlbumArtURIs(new URI[] { getDispatcher().getAlbumProcessor().createAlbumArtURI(album) });
+            if (album.getCoverArtPath() != null) {
+                container.addProperty(factory.toAlbumArt(album));
+            }
```
tesshucom, line 141: This conditional expression is not necessary.

tesshucom: Done.

Issue #2349, opened 2023-09-05
Rebuild menu items for Web, UPnP


tesshucom:
Prerequisites: #1879

This issue has a lot of work to do, so it will likely span several versions until it is finally resolved. Related subtasks and pull requests will be created accordingly.

### Overview

 - Web Page menu items will be renewed
   - Left page (drawer) items will be recreated anew
   - Users can enable only the items they need from the settings page. Sorting and renaming will be also supported
 - UPnP Top nodes will provide similar functionality. Similar functionality will be provided, except for technically unacceptable.

### Problem description

The web page hasn't changed much since Subsonic/Airsonuc. As a result, traditionally recurring complaints about the UI have not been resolved much. So how should we solve them. Well... the menu should be in place first and foremost. When you enter a restaurant, the first thing you do is look at the menu.

Of course, the ingredients and seasonings included in each dish are also important. But it all starts with the menu. The menu must include dishes that the restaurant offers. It's not like you can just order sushi at a prestigious French restaurant. 

### Current situation

 - Subsonic had a decent UI for its time. At least the era before that was much less sane, and the era before that didn't even have a UI. We can show some respect for that. Several improvements were made during the Air Sonic era.
 - Most of UPnP was developed after Jpsonic. 
   - It remains the most versatile protocol for performing media-related communications. It is impossible to achieve sufficient Player functionality using a browser. To get a fairly good **playback** experience, you'll need to use Supsonic's client or the more advanced UPnP Player.
   - Items similar to or complementary to the original Web Page Items were added. For example, ID3 browsing.
 - There is something that somehow seems like that. However, it is quite chaotic.

Let's roughly categorize them. Items that exist in General classification are likely to exist as menu items in Music Player.

![image](https://github.com/tesshucom/jpsonic/assets/27724847/f942eae7-ecbd-43c9-b95b-39364efe998f)

 - What is said to be IMPOSSIBLE does not mean abandoning possibilities. For example, if you are using it completely alone, there is no need to worry about account conflicts. On a limited basis, it is possible to provide these by specifying an account.
 - It doesn't mean abandoning what isn't here. It just shows the status quo. For example, the first thing that comes to mind is that Rating-browsing doesn't exist. It is not important right now to propose or announce such things. The problem is that there is no design that allows systematic management.

### Goal

It will likely span several versions until it is finally resolved.

 - Create a setting page that can control web page and UPnP menu
 - Link existing features and new menu settings
[...]

PR #2441, opened 2023-11-07
Fix getAlbumList2 sorting to ignore case

tesshucom:
### Overview

Fix for getAlbumList2 of Subsonic API . If alphabeticalByName or alphabeticalByArtist is used in API Options, it will be fixed to be case insensitive.

### Details

As for internal processing, byArtist/ignoreCase is defined as an argument of the Dao method.

https://github.com/airsonic/airsonic/blob/5ccca059d5cfe3dd19734c27861b434ea21b43d8/airsonic-main/src/main/java/org/airsonic/player/controller/SubsonicRESTController.java#L1136-L1138

 - Letter case issue often comes up on the agenda even on general media servers. It is a common opinion that uppercase and lowercase letters are ignored in Alphabetical's natural sort.
 - Traditional servers avoided case-insensitive sorting on the database side for performance reasons. (Perhaps. A large number of string conversions were avoided.)

Therefore, it is thought that there is no major problem in changing this specification.

### Note1
getAlbumList2 (get all ID3 albums) will be the change with the greatest impact among these sort-fixes. ID3 has a normalized table, so in the case of Album, a join may be necessary depending on the conditions. 

 - **1** --- If it is only Album, it will be sorted by album_order
 - **2** --- If you want to sort albums by artist and album, they will be sorted by artist_order, album_order. Therefore we will need JOIN

Please note the following points.

 - In the current Jpsonic, only **1** is used. It's used in UPnP album view.
 - The album list on the web page is sorted by Artist -> Album. However, this is a File Structure (media_file table). No JOIN.

Performance verification will be required when replacing Album All View on the Wave page with ID3. Related #2349.

### Note2

The Dao design will be maintained internally for some time. However, implementations related to Letter case in Dao may be removed at an appropriate stage.

 - I believe that the legacy server implementation was never intended to have a case-sensitive use case and is simply a result of technical constraints.
 - There is no point in sticking only to case-sensitive. 
   - This is because we need many other perspectives for sort by name. Advanced natural sort of strings is not possible on Database.

### DIFF
--- a/jpsonic-main/src/test/java/com/tesshu/jpsonic/dao/AlbumDaoTest.java
+++ b/jpsonic-main/src/test/java/com/tesshu/jpsonic/dao/AlbumDaoTest.java
@@ -1,0 +1,123 @@
+/*
+ * This file is part of Jpsonic.
+ *
+ * Jpsonic is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Jpsonic is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * (C) 2023 tesshucom
+ */
+
+package com.tesshu.jpsonic.dao;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.lang.annotation.Documented;
+import java.util.Collections;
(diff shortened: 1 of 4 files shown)


jpsonic-main/src/test/java/com/tesshu/jpsonic/dao/AlbumDaoTest.java
```diff
@@ -96,28 +96,28 @@ private ArgumentCaptor<String> getQuery(boolean byArtist, boolean ignoreCase) {
         @AlphabeticalOps.Conditions.IgnoreCase.True
         @Test
         void c00() {
-            assertEquals("LOWER(artist_reading), album_order, LOWER(name_reading)", getOrderBy(getQuery(true, true)));
+            assertEquals("artist_order, album_order", getOrderBy(getQuery(true, true)));
```
tesshucom, line 99: Modify so that Join is performed only under this condition

tesshucom: Done.

PR #2446, opened 2023-11-15
UPnP implementation improvements

tesshucom:
Prerequisites: #2436

### Overview

UPnP will be reimplemeted. That has two purposes.

 - Music index caching mechanism will be removed in this Pull request
   -  Related #2431
   - Most of the processing for index generation can now be done using SQL instead of Java, making it faster. There are many benefits to eliminating complicated caches.
 - Implementation modifications due to design improvements
   -  Related #2436. In this pull request, I will try to clean up the existing code. A major overhaul of the settings screen is also planned in the near future. #2349 If these things go well, it will be easier for new Views to enter. New genre-view, Year, Movie, Composer ... things like that.

Designing and implementing UPnP requires more rigorous planning than web pages or Rest. If UPnP does these things well, it is expected that it will be **highly reusable for other protocols and UPnP-derived protocols**.

The SQL used in UPnP is characterized by more variations in parameters, and will basically be upwardly compatible with Web and Rest.

 -  SORT : In most cases the order columns are used because speed is required. (Java Comparator is not used) 
 -  COUNT/LIMIT : Most lists come with a count/limit. Most UPnP apps seem to take advantage of this.
 - MEDIA TYPE : In UPnP, media types are clearly filtered for each tree node. (There are some ambiguous areas in Web and Rest.)

### Details

The process performed by this pull request can be divided into three parts as follows:

 - [x] Implementation changes due to new design (Described later)
   - Since the amount is large, commit merge and review. 
 - [x] Organize MediaFileDao queries
   - Delete unnecessary queries
   - Combine similar queries
   - Define filters for MediaType better
 - [x] Remove Ehcache implementation for MusicIndex. ...This was the original purpose 🙄 
 - [x] Change class name
   - Following Airsonic's practice, the suffix UpnpProcessor is often used. Too long. It looks like Proc is fine since it is in Package. It's not a big deal. However, since `@Qualifier` changes in Spring, it is best to do this just before the final operation check.

<details>
<summary>Implementation changes due to new design</summary>

In short, it's what happens under the hood. It's not that big a deal. The commit will be heavily 'squash' as the difference in code is large compared to the difficulty of the content.

 - Clarification of entities used in tree representations
   - This is naturally achieved by implementing the newly created interface UPnPContentProcessor.
 - Improving test coverage
   - Tests will be added to most classes. The order of class methods and test methods has been unified to be the same as UPnPContentProcessor. (Therefore, there are many Diff 🙄 )
 - Add CountLimitProc
   - One of the causes of complexity is the calculation of count values. They were made common by interface CountLimitProc.
   - In UPnP, most communications are executed with  count/limit. Jpsonic has faithfully implemented this, so you will be able to browse large libraries very comfortably. However, some Processors require a limit to be placed on the server side for requests. `Recently`, `Shuffle`, `Search results`, etc. 
 - Add UpnpDIDLFactory
   - The process of converting server entities to DIDL has been standardized. Airsonic had a somewhat idiosyncratic implementation, so those were removed. Item like 'ALL'. Hmm, Twonky may have had such an implementation 🤔 That may be a big disadvantage depending on the Client App. ... No, it will be.

</details>

### Goal

 - Delete the UPnP service cache and make it completely stateless
 - Organize the actual equipment to make it easier to add functions in the future. To avoid degrade, some of the SQL used in UPnP had been written separately from the Web and Rest implementations. These will be gradually integrated. 
 - Improve code coverage

### Non-Goal

Our future goal is to extend the redesign being done in the UPnP implementation to Rest and web pages. (Eliminating Java side logic and doing what can be done with SQL with SQL)

However, this pull request doesn't fix all of them. Web page and REST modifications are avoided in this pull request.

### Impact on users

There is no particular problem since most of the changes are View changes. However, the processing that was missing in #2423 is included in this commit. (Case of creating Music-Index for Album directly under Root)

 - When looking at this pull request alone, a full scan may be required for libraries with special directory structures.
 - However, the next release will be a major version upgrade. A full scan will be forced because there is a schema update in the Database. (In other words, there is no need to be conscious of the details.)

This commit includes File Structure enhancements that allow browsing of rather strange directory structures. For example, a test case has been added that scans the file structure shown below. 

image

In this tree, Album1 has been corrected to display correctly. It's a rare case though!


tesshucom:
Hmm, it looks like the next fix hasn't been started yet. 

 - WMPProcessor.java
 - PlaylistUpnpProcessor.java
 - PodcastUpnpProcessor.java

It's not a relatively difficult classe, so let's finish it quickly. There are no plans to change WMPProcessor anymore. (for Windows Medai Player only) . Therefore, you can skip WMPProcessor. Once these are completed, code coverage will be complete. There are too many commits, so it would be a good idea to crush it once before reviewing it.

tesshucom:
There is a dependency on the order of each commit (because some classes have inheritance changed). Since there are too many commits, they will be categorized into 4 groups and squashed.

jpsonic-main/src/main/java/com/tesshu/jpsonic/dao/MediaFileDao.java
```diff
@@ -140,6 +144,37 @@ public List<MediaFile> getChildrenWithOrderOf(String path) {
...
+    }
+
+    public List<MediaFile> getArtists(MusicIndex musicIndex, List<MusicFolder> folders, long offset, long count) {
+        return template.query("select " + QUERY_COLUMNS + """
+                from artist
+                where music_index=?
+                order by artist_order
+                offset ? limit ?
+                """, rowMapper, musicIndex.getIndex(), offset, count);
+    }
```
tesshucom, line 176: In this method, only directory, album, and music can be included in the result. This should be redesigned...　Instead of rewriting the SQL, you may be able to share it with other logic by adding it to the filter condition in the method Paramater.

Well, in the case of a web page, for example, if you mix video and music in the same directory, I think only the music will be displayed. (Although it's a little vague) If there was only a video, it probably would have jumped to a dedicated View. 

Browsing must be designed so that there are basically no omissions. However, it is standard for media-related software to generate different tree nodes for each media type. (Otherwise, when you add a directory to the queue, music and videos will be mixed together in the queue.)

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/MediaFileService.java
```diff
@@ -170,6 +172,11 @@ public List<MediaFile> getChildrenWithoutSortOf(MediaFile parent, boolean includ
         return result;
     }
 
+    public List<MediaFile> getDirectChildren(MusicIndex musicIndex, List<MusicFolder> folders, long offset,
+            long count) {
+        return mediaFileDao.getDirectChildren(musicIndex, folders, offset, count);
+    }
+
```
tesshucom, line 179: This would require a redesign, similar to the Dao method.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/AlbumUpnpProcessor.java
```diff
@@ -57,49 +56,34 @@ public ProcId getProcId() {
...
 
-    @Override
-    public int getDirectChildrenCount() {
-        return albumDao.getAlbumCount(util.getGuestFolders());
-    }
-
     @Override
     public List<Album> getDirectChildren(long offset, long maxResults) {
         return albumDao.getAlphabeticalAlbums((int) offset, (int) maxResults, false, true, util.getGuestFolders());
     }
```
tesshucom, line 67: Confirm that it is Alphabetical only here. (Other Album view will be branched to Alphabetical or Year depending on settings)

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/IndexId3UpnpProcessor.java
```diff
@@ -19,406 +19,72 @@
...
 
     @Override
-    public void addItem(DIDLContent didl, Id3Wrapper item) {
-        if (!item.isSong() || item.isIndex()) {
-            didl.addContainer(createContainer(item));
-        } else {
-            didl.addItem(factory.toMusicTrack(item.getSong()));
-        }
+    public List<MusicIndex> getDirectChildren(long offset, long maxLength) {
+        return musicIndexService.getIndexedId3ArtistCounts(util.getGuestFolders()).keySet().stream().toList();
```
tesshucom, line 62: Use offset and count.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/MediaFileUpnpProcessor.java
```diff
@@ -62,72 +59,45 @@ public ProcId getProcId() {
...
+    public List<MediaFile> getDirectChildren(long fastResult, long maxResults) {
+        List<MusicFolder> folders = util.getGuestFolders();
+        if (folders.isEmpty()) {
+            return Collections.emptyList();
+        } else if (folders.size() == SINGLE_MUSIC_FOLDER) {
+            MediaFile folder = mediaFileService.getMediaFile(folders.get(0).toPath());
+            return getChildren(folder, fastResult, maxResults);
         }
-        return count;
+        return folders.stream().map(folder -> mediaFileService.getMediaFile(folder.toPath())).toList();
```
tesshucom, line 91: Use offset and count.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/UpnpDIDLFactory.java
```diff
@@ -63,6 +75,9 @@
 @Component
 public class UpnpDIDLFactory implements CoverArtPresentation {
 
+    private static final ThreadLocal<DateTimeFormatter> DATE_FORMAT = ThreadLocal
+            .withInitial(() -> DateTimeFormatter.ofPattern("yyyy-MM-dd").withZone(ZoneId.systemDefault()));
+
```
tesshucom, line 80: SUB_DIR_EXT : private static final is on top.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/UpnpDIDLFactory.java
```diff
@@ -183,6 +198,63 @@ private MimeType getMimeType(MediaFile song, Player player) {
...
+        container.addProperty(toPlaylistArt(playlist));
+        return container;
+    }
+
+    public MusicArtist toArtist(MediaFile artist, int childCount) {
+        MusicArtist container = new MusicArtist();
+        container.setTitle(artist.getName());
+        container.setId(ProcId.FOLDER.getValue() + ProcId.CID_SEPA + artist.getId());
+        mediaFileService.getParent(artist).ifPresent(parent -> container.setParentID(String.valueOf(parent.getId())));
+        artist.getCoverArtPath().ifPresent(path -> container.addProperty(toArtistArt(artist)));
```
tesshucom, line 253: All methods that generate cover art can be changed to private.

 - toArtistArt
 - toAlbumArt
 - toPodcastArt
 - toPlaylistArt

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/MediaFileUpnpProcessor.java
```diff
@@ -62,72 +59,45 @@ public ProcId getProcId() {
...
-            }
-            MusicArtist container = new MusicArtist();
-            applyId(item, container);
-            item.getCoverArtPath().ifPresent(path -> container.addProperty(factory.toAlbumArt(item)));
-            return container;
+    public void addItem(DIDLContent parent, MediaFile entity) {
+        if (entity.isFile()) {
+            parent.addItem(factory.toMusicTrack(entity));
+        } else {
+            parent.addContainer(createContainer(entity));
```
tesshucom, line 78: The name addItem is inappropriate. (The interface is not good) The pattern of adding either Container or Item is often seen.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/AlbumByGenreUpnpProcessor.java
```diff
@@ -60,100 +52,17 @@ public ProcId getProcId() {
...
 
     @Override
-    public void addChild(DIDLContent didl, MediaFile song) {
-        if (isEmpty(song.getMediaType())) {
-            didl.addItem(factory.toMusicTrack(song));
-        } else {
-            didl.addContainer(createContainer(song));
-        }
+    public void addChild(DIDLContent didl, MediaFile album) {
+        didl.addContainer(factory.toAlbum(album, mediaFileService.getChildSizeOf(album)));
```
tesshucom, line 66: The name is confusing. 

> DIDLContent didl

In this case `parent` is better. Fix not just here but overall.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/AlbumUpnpProcessor.java
tesshucom: No problem.

[Alphabetical]
AlbumUpnpProcessor

[Alphabetical/Year]
ArtistUpnpProcessor
ArtistByFolderUpnpProcessor
MediaFileUpnpProcessor
AlbumByGenreUpnpProcessor -> MediaFileUpnpProcessor
IndexUpnpProcessor -> MediaFileUpnpProcessor

[Random]
RandomAlbum

[Created]
RecentAlbumId3UpnpProcessor
RecentAlbumUpnpProcessor

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/PlaylistUpnpProcessor.java
```diff
@@ -51,25 +50,18 @@ public ProcId getProcId() {
...
-        return container;
+    public Container createContainer(Playlist playlist) {
+        return factory.toPlaylist(playlist);
     }
 
     @Override
-    public int getDirectChildrenCount() {
-        return playlistService.getCountAll();
+    public List<Playlist> getDirectChildren(long offset, long maxResults) {
+        return PlayerUtils.subList(playlistService.getAllPlaylists(), offset, maxResults);
```
tesshucom, line 59: Don't use PlayerUtils.subList. (Remnants of old Java code)

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/IndexId3UpnpProcessor.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/MediaFileUpnpProcessor.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/UpnpDIDLFactory.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/UpnpDIDLFactory.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/PlaylistUpnpProcessor.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/AlbumByGenreUpnpProcessor.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/MediaFileUpnpProcessor.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/dao/MediaFileDao.java
tesshucom: #### Methods to be fixed

- MediaFileUpnpProcessor#getChildren
- MediaFileUpnpProcessor#getChildSizeOf
- MediaFileDao#getSingleSongs
- IndexUpnpProcessor#getChildren
- IndexUpnpProcessor#getChildSizeOf

#### Problem

We need to filter the Mediatype accurately when expanding the directory. For the above method:

 - Permissible : MediaType.MUSIC, MediaType.DIRECTORY, MediaType.ALBUM
 - Not acceptable : MediaType.PODCAST, MediaType.AUDIOBOOK, MediaType.VIDEO

This is because it is a tree node for handling music. Users browse the directory and often play the entire directory. At that time, media other than music must not be mixed. PODCAST, AUDIOBOOK, VIDEO should each be provided with a different tree node. (PODCAST is already implemented. AUDIOBOOK, VIDEO will also be created in the future) These are not specifications limited to Jpsonic in particular. Many UPnP media servers provide such browsing for user convenience.

#### Approach

A new method will be created that takes the MediaType filter as an argument, taking into account offset/count/order like the existing method. It is currently necessary, and is expected to be necessary at least for browsing AUDIOBOOK and VIDEO on UPnP.

We are currently considering web pages. (The design is different from UPnP) It won't be addressed right away. But personally I feel that it would be easier to understand if AUDIOBOOK and VIDEO browsing were isolated like UPnP. The view design of Subsonic is mixed type.

jpsonic-main/src/test/java/com/tesshu/jpsonic/dao/MediaFileDaoTest.java
```diff
@@ -26,613 +26,702 @@
...
-            assertTrue(query.endsWith(ORDER_LIMIT));
-            assertTrue(suffix.length() < query.length());
+        void testGetMudicIndexCounts() {
+            List<MusicFolder> folders = List.of(MUSIC_FOLDER);
+            List<IndexWithCount> counts = mediaFileDao.getMudicIndexCounts(folders, Collections.emptyList());
+            assertEquals(2, counts.size());
+            counts.stream().forEach(index -> {
+                switch (index.index()) {
+                case "D" -> assertEquals(2, index.artistCount());
+                case "" -> assertEquals(1, index.artistCount());
```
tesshucom, line 721: This case is not good.  It has not been decided how to index the Album directory that exists directly under the music folder. (That's a very irregular case though.)

tesshucom: Therefore, the following modifications are necessary.

 - In the above test case, modify so that Album directly under MusicFolder is indexed.
 - Fix the field name of Record used in MediaFileDao

jpsonic-main/src/main/java/com/tesshu/jpsonic/dao/MediaFileDao.java
```diff
@@ -1158,4 +1191,7 @@ Optional<String> getShowUnstarredSongsCondition() {
             return Optional.empty();
         }
     }
+
+    public record IndexWithCount(String index, int artistCount) {
+    }
```
tesshucom, line 1196: count or directoryCount is better instead of artistCount. Album is also counted.

jpsonic-main/src/test/java/com/tesshu/jpsonic/dao/MediaFileDaoTest.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/dao/MediaFileDao.java

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/scanner/ScannerProcedureService.java
```diff
@@ -815,6 +815,30 @@ boolean updateSortOfArtist(@NonNull Instant scanDate) {
...
+     * https://github.com/tesshucom/jpsonic/pull/2446#discussion_r1403505056
+     */
+    private void updateIndexOfAlbum() {
+        List<MusicFolder> folders = musicFolderService.getAllMusicFolders();
+        if (mediaFileDao.getChildSizeOf(folders, MediaType.ALBUM) == 0) {
+            return;
+        }
+        folders.stream().forEach(folder -> {
+            int offset = 0;
+            List<MediaFile> albums = mediaFileDao.getChildrenOf(folder.getPathString(), offset, ACQUISITION_MAX, false);
```
tesshucom, line 829: Filters will be implemented later

tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/MediaFileService.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/dao/MediaFileDao.java

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/UpnpDIDLFactory.java
```diff
@@ -183,6 +198,63 @@ private MimeType getMimeType(MediaFile song, Player player) {
         return mimeTypeString == null ? null : MimeType.valueOf(mimeTypeString);
     }
 
+    private StorageFolder createMusicFolder(int id, String name, ProcId procId, int childCount) {
```
tesshucom, line 201: So far. . . There seems to be no problem with data flow. However, there may be a problem with image output. Confirmation required.

tesshucom: ah. . This is my misunderstanding. When checking on the actual device, I made a mistake in setting the Base LAN URL. 🤫🤫🤫

Issue #2474, opened 2023-12-04
Scan stalls when going into a folder with three dots


Tak-MK:
## Problem description

Hi there!
The gist is that everytime jpsonic tries to do the step `[12/22]REFRESH_ALBUM_ID3` / `UPDATE_ORDER_OF_SONG` inside of a folder with three dots, it stalls with this log:
```
java.lang.SecurityException: Access denied to file : /jpsonic/vgm/Shin Megami Tensei/Shin Megami Tensei If.../PS1
        at com.tesshu.jpsonic.service.MediaFileService.getMediaFile(MediaFileService.java:101) ~[classes!/:112.2.11]
        at com.tesshu.jpsonic.service.MediaFileService.getParentOf(MediaFileService.java:139) ~[classes!/:112.2.11]
        at com.tesshu.jpsonic.service.MediaFileService.getParent(MediaFileService.java:143) ~[classes!/:112.2.11]
        at com.tesshu.jpsonic.service.scanner.ScannerProcedureService.albumId3Of(ScannerProcedureService.java:534) ~[classes!/:112.2.11]
        at com.tesshu.jpsonic.service.scanner.ScannerProcedureService.lambda$createAlbumId3s$16(ScannerProcedureService.java:586) ~[classes!/:112.2.11]
        at java.base/java.util.Optional.ifPresent(Unknown Source) ~[na:na]
        at com.tesshu.jpsonic.service.scanner.ScannerProcedureService.createAlbumId3s(ScannerProcedureService.java:585) ~[classes!/:112.2.11]
        at com.tesshu.jpsonic.service.scanner.ScannerProcedureService.refleshAlbumId3(ScannerProcedureService.java:605) ~[classes!/:112.2.11]
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(Unknown Source) ~[na:na]
        at java.base/java.lang.reflect.Method.invoke(Unknown Source) ~[na:na]
        at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:344) ~[spring-aop-5.3.30.jar!/:5.3.30]
        at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:198) ~[spring-aop-5.3.30.jar!/:5.3.30]
[...]

PR #2492, opened 2023-12-09
Fix to avoid handling files or directories ending with a dot

tesshucom:
## Overview

Add processing for cases where the file name or directory name ends with a dot.

 - Files or directories ending with a dot will be modified to not be scanned
 - When outputting a file locally using a title such as m3u or Podcast, the Dot at the end will be modified to be deleted.
 - Extra : Some reserved words used in the latest Synology DSM will be added

## Problem description

Jpsonic will immediately stop processing if an unexpected case is detected in the scan.

 - #2474

Currently, the scan will stop if you try to scan a file or directory that ends with a dot.

## Background

Files or directories ending with a dot are prohibited, especially on Windows Explorer. (It cannot be created from Explorer. Also, directories that end with Dot cannot be expanded. Both will result in an error.) However, they can be created from cmd or Program.

General Windows apps often take care not to create such files and directories. For example, when creating a directory from the Album name, change the dot to another string such as an underscore. One of historical classical trivialities that often require workarounds on Windows, such as "filenames with spaces". When transferring files created with ripping software without such specifications or Linux-specific software to Windows, cross-platform functionality may be hindered.

Conversely, it is a case that is not often encountered, so it may be said that it was not taken care of on the Sonic server.

## Policy on this issue

Subsonic/Airsonic originally has a definition of "file that cannot be read". This will be extended to exclude files or directories that end with Dot.

 https://github.com/airsonic/airsonic/blob/5ccca059d5cfe3dd19734c27861b434ea21b43d8/airsonic-main/src/main/java/org/airsonic/player/service/MediaFileService.java#L458-L471

 Also, the existing Synology reserved words will be updated. 

 > https://kb.synology.com/en-in/DSM/help/FileStation/connect?version=7

 >  - The system cannot display files/folders that adopt the following naming formats:
 >    - Beginning with ._ or .SYNOPPSDB
 >    - Including /
 >    - Named exactly as .    ..    '    "    .DS_Store    `@eaDir`    `@sharebin`    `@tmp`    .SynologyWorkingDirectory

### Specifications regarding warnings when excluded scanning

It is not always helpful to output information about all excluded files to the log. (If anything, verbose logs are often detrimental to the System!) Therefore, Jpsonic's policy is to suppress log output as much as possible for obvious problems. Jpsonic's default Log Level is WARN. Therefore, information about most excluded files will not be output to the log, as shown in the table below. (The log level can be changed using the startup argument.)

Prohibition rules | Log Level
-- | --
Thumbs.db (Tiny image created by Windows) | No warning
SYNOLOGY reserved word | No warning
SymbolicLink | INFO
ExcludePattern specified by the user on the settings page | INFO
Starts with double Dot | INFO
**Ends with Dot** | **WARN**

**Note that the spec added in this commit will be warned with WARN level.** The reason is as follows.

 - Basically, it is assumed that this is a rare case.
 - If situation to this case occur, there is a high possibility that the user was not aware of the problem.
   - The ripping software will properly create file and directory names from the title. Most of the time we take it for granted.
   - In case that wasn't the case. If Jpsonic uses it silently, the problem could be masked.
   - This kind of thinking would be called ``Fail Fast``

### Best Practice if you run into this problem

It's not necessarily a good idea for a program that loads a file to automatically rewrite its Path to read or display it. That's because it's a stopgap measure and not a fundamental solution. It must be fail-fast to work securely with many platforms and with many other software.

- Please note that there are platform-dependent restrictions on the strings that can be used within the file system.
- In such cases, workaround is to rename the target directory/files to UNICODE. This is not an unusual task in non-Latin-speaking countries. Unless you are using a very antique OS, it will probably work. 
  - `… (Horizontal Ellipsis)` `․ (one dot leader)`
  - Apostrophe...hmm. Maybe you should ask a French engineer.
    - [Which Unicode character should represent](https://tedclancy.wordpress.com/2015/06/03/which-unicode-character-should-represent-the-english-apostrophe-and-why-the-unicode-committee-is-very-wrong/)

How it looks depends on the font. Some fonts may even be visually indistinguishable.

image

### Todo

 - [x] 1. Fix file/dir exclusion rules when scanning
 - [x] 2. Check file/dir generation rules in Podcast/m3u
   - It had been be considered. StringUtil.fileSystemSafe is used to force change to hyphen. 
   - But it's a shame. A greedy transformation should be done here. The trailing dot will be fixed to be removed.
 - [x] 3. Check the Spec of Download/Zip Download 
   - There is no problem with zip files because the extension ".zip" is added.
   - There is no problem with processing within zip Stream (directory structure inside zip) as long as (1) is guaranteed.
 - [x] 4. Confirmation is required regarding the judgment of cover art.
   - Sonic servers support embedded cover art and local cover art. File validation for embedded cover art(=Mediafile) will be covered by (1). Other files will require Dot detection before determining whether they are image files. (Otherwise we will get a NIO Error)

### DIFF
--- a/jpsonic-main/src/main/java/com/tesshu/jpsonic/service/SecurityService.java
+++ b/jpsonic-main/src/main/java/com/tesshu/jpsonic/service/SecurityService.java
@@ -66,6 +66,10 @@ public class SecurityService implements UserDetailsService {
 
     private static final Logger LOG = LoggerFactory.getLogger(SecurityService.class);
 
+    // https://kb.synology.com/en-in/DSM/help/FileStation/connect?version=7
+    private static final List<String> SYNOLOGY_RESERVED_WORDS = List.of("._", ".SYNOPPSDB", ".DS_Store", "@eaDir",
+            "@sharebin", "@tmp", ".SynologyWorkingDirectory");
+
     private final UserDao userDao;
     private final SettingsService settingsService;
     private final MusicFolderService musicFolderService;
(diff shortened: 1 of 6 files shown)


jpsonic-main/src/main/java/com/tesshu/jpsonic/service/SecurityService.java
```diff
@@ -525,28 +529,61 @@ protected boolean isFileInFolder(final String file, final String folder) {
...
+        // Exclude Thumbnail on Windows
+        if ("Thumbs.db".equals(name)) {
             return true;
         }
 
-        // Exclude all hidden files starting with a single "." or "@eaDir" (thumbnail dir created on Synology devices).
-        return !name.isEmpty() && name.charAt(0) == '.' && !name.startsWith("..") || name.startsWith("@eaDir")
-                || name.startsWith("@tmp") || "Thumbs.db".equals(name);
+        // Exclude files or dir created on Synology devices
+        return SYNOLOGY_RESERVED_WORDS.stream().anyMatch(name::equals);
```
tesshucom, line 587: I noticed it after merging. Strictly speaking, these words need to be classified by partial match and prefix match. However, the result does not change because the word with a prefix match is unreachable due to the previous "Dot at the beginning" rule. It will be released as is since there is no impact on processing. This may be fixed in a future refactoring.

PR #2541, opened 2024-01-15
Migrate cling to jupnp

tesshucom:
## Overview

Cling, the library used for UPnP, will be migrated to Jupnp. Cling is already EOL. Jupnp is a derivative of Cling and is currently undergoing maintenance.

 - https://github.com/4thline/cling
   - A library that has existed almost from the beginning of Java UPnP history. An example is BubbleUPnP. (I don't know what BubbleUPnP uses at this point)
 - https://github.com/jupnp/jupnp
   - A derivative of Cling. Adopted by Universal Media Server.

This will have the following impact on product configuration:

 - The number of CVE suppression settings will be 0. In short, it's Perfect.
 - Cling related libraries will be removed from the local Repository
   - Like many Java Projects, the required libraries are obtained from Maven Central. However, in a Java Project, a static library may be maintained locally in the Project if it is absolutely necessary during a transition period of the library. In Jpsonic, Cling was managed in a local Repository.
   - As a result, only one library is managed by the local Repository. This library is also scheduled to be removed due to modernization. 

## Details

Library changes and associated refactorings.

 - Migrate cling to jupnp 
 - Remove cling and cling related libs from local repository
 - Refactoring : Change upnp entity package
 - Refactoring: Change package of some upnp classes
 - Add UpnpServiceFactory 
 - Remove unnecessary suppression

Unlike previous implementations, accepting UPnP queries will use a Java standard simple http server rather than a socket implementation. Along with this, tuning using some parameters is possible. Links for their default values and detailed information have been added as comments to production.yml. 

 - Add UPnP-params to Dockerfile 

Unlike previous implementations, Classes provided as standard by the library are not used for thread pool and thread factory. It has been replaced to using Spring's thread management mechanism via an adapter. The use of default implementation was be avoided in order to meet requirements such as shutsown being linked to the application lifecycle and shutdown flow, ensuring logging of UncaughtExceptions, and not accessing the SecurityManager.

 - Add UpnpServiceConfigurationAdapter


tesshucom:
Since production.yml has been changed, an announcement is required at the time of release. It is backward compatible, so you can continue to use previous versions of yml as is.

PR #2564, opened 2024-01-26
Migrate to ReentrantLock

tesshucom:
Prerequisites: #2558

## Overview

The synchronized codings will be removed and replaced with concurrent Lock classes. 

This is part of complying with [Practical Advice for Virtual Threads](https://docs.oracle.com/en/java/javase/20/core/virtual-threads.html#GUID-9065C2D5-9006-4F1A-93E0-D5153BB40475). 

## Goal

synchronize will be replaced by ReentrantLock or ReentrantReadWriteLock. 

## Non-Goal

 - In particular, it does not include thread modifications.
 - Only changes to CoverArtController are not included in this pull request. CoverArtController has Semaphores intentionally killed. This is because there was a problem with the synchronization process. Currently, Jpsonic should be free from glitches such as black images, incorrect images, and updates not being reflected, in exchange for image processing being single-threaded. CoverArtController will be refactored in a subsequent pull request.

## Details

It will be mainly divided into three groups. These will be "squash" after review

- Remove unnecessary synchronized
- Migrate to ReentrantLock
- Migrate to ReentrantReadWriteLock

Jpsonic's synchronized code has the following history.

- Eliminating synchronized methods
  - They have been replaced by synchronized blocks using object locks. There was a lot of misuse of synchronized methods, and some of them didn't make sense.
- Scrutiny by Facebook Infer
  - A tool called RacerD is excellent. Since I set these to no errors once, I think that many fatal problems have been avoided.

However, the points pointed out by the tool were corrected very mechanically. (Unnecessary synchronized or misunderstood codes cannot be detected, and new ones are sometimes created.) This time, unnecessary synchronized files are deleted manually.

### DIFF
--- a/airsonic-taglibs/src/main/java/com/tesshu/jpsonic/util/StringUtilBase.java
+++ b/airsonic-taglibs/src/main/java/com/tesshu/jpsonic/util/StringUtilBase.java
@@ -39,7 +39,6 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 public final class StringUtilBase {
 
     private static final long BINARY_1KB = 1024L;
-    public static final Object FORMAT_LOCK = new Object();
 
     /**
      * Disallow external instantiation.
@@ -64,33 +63,31 @@ public final class StringUtilBase {
      * @return The formatted string.
      */
     public static String formatBytes(long byteCount, Locale locale) {
-        synchronized (FORMAT_LOCK) {
-            // More than 1 TB?
-            if (byteCount >= BINARY_1KB * 1024 * 1024 * 1024) {
-                NumberFormat teraByteFormat = new DecimalFormat("0.00 TB", new DecimalFormatSymbols(locale));
-                return teraByteFormat.format(byteCount / ((double) 1024 * 1024 * 1024 * 1024));
-            }
-
-            // More than 1 GB?
-            if (byteCount >= BINARY_1KB * 1024 * 1024) {
-                NumberFormat gigaByteFormat = new DecimalFormat("0.00 GB", new DecimalFormatSymbols(locale));
-                return gigaByteFormat.format(byteCount / ((double) 1024 * 1024 * 1024));
-            }
-
-            // More than 1 MB?
-            if (byteCount >= BINARY_1KB * 1024) {
-                NumberFormat megaByteFormat = new DecimalFormat("0.0 MB", new DecimalFormatSymbols(locale));
-                return megaByteFormat.format(byteCount / ((double) 1024 * 1024));
-            }
-
-            // More than 1 KB?
-            if (byteCount >= BINARY_1KB) {
-                NumberFormat kiloByteFormat = new DecimalFormat("0 KB", new DecimalFormatSymbols(locale));
-                return kiloByteFormat.format((double) byteCount / 1024);
-            }
-
-            return byteCount + " B";
+        // More than 1 TB?
+        if (byteCount >= BINARY_1KB * 1024 * 1024 * 1024) {
+            NumberFormat teraByteFormat = new DecimalFormat("0.00 TB", new DecimalFormatSymbols(locale));
+            return teraByteFormat.format(byteCount / ((double) 1024 * 1024 * 1024 * 1024));
         }
+
+        // More than 1 GB?
+        if (byteCount >= BINARY_1KB * 1024 * 1024) {
+            NumberFormat gigaByteFormat = new DecimalFormat("0.00 GB", new DecimalFormatSymbols(locale));
+            return gigaByteFormat.format(byteCount / ((double) 1024 * 1024 * 1024));
+        }
+
+        // More than 1 MB?
+        if (byteCount >= BINARY_1KB * 1024) {
+            NumberFormat megaByteFormat = new DecimalFormat("0.0 MB", new DecimalFormatSymbols(locale));
+            return megaByteFormat.format(byteCount / ((double) 1024 * 1024));
+        }
+
+        // More than 1 KB?
+        if (byteCount >= BINARY_1KB) {
+            NumberFormat kiloByteFormat = new DecimalFormat("0 KB", new DecimalFormatSymbols(locale));
+            return kiloByteFormat.format((double) byteCount / 1024);
+        }
+
+        return byteCount + " B";
     }
 
     /**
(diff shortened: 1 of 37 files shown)


tesshucom:
Although it is correct as a process, it may affect the test. There is a test in which operation is confirmed by dynamically changing the home directory 🤒
SettingsServiceTest
LegacyHsqlUtilTest

jpsonic-main/src/main/java/com/tesshu/jpsonic/domain/PlayQueue.java
```diff
@@ -94,24 +95,28 @@ public void setName(String name) {
...
                     // Remove file from playlist if it doesn't exist.
                     if (!file.exists()) {
                         files.remove(index);
                         index = Math.max(0, Math.min(index, size() - 1));
                         return getCurrentFile();
                     }
-
                     return file;
+                } finally {
+                    sequenceLock.writeLock().unlock();
```
tesshucom, line 115: Use ReadWriteLockSupport

tesshucom:
> プロセスとしては正しいですが、テストに影響を与える可能性があります。ホームディレクトリを動的に変更して動作を確認するテストあります🤒 SettingsServiceTest LegacyHsqlUtilTest

The motivation is to improve the performance of how to obtain the home directory. Until now, it was validating the directory every time it was accessed. However, this directory typically does not change, so it may be sufficient to verify it only once after startup. (Redundant locks to ensure unnecessary IO access and IO access. By the way, this path is used very frequently.)

The problem is that if you make it static, while it contributes to performance, tests that change the home directory will no longer work. (Removing these tests would be unacceptable.) The elegant method is to control it on the junit side, but these methods have changed in various ways depending on the version, and currently it is very complicated to handle due to the specifications of the security manager.

So the easiest fix right now is to use the long-existing isDevelopmentMode flag. But this isDevelopmentMode also has a problem

`System.getProperty("airsonic.development")`

If this were included in production code it would degrade performance. Therefore, a new setter will be added for isDevelopmentMode. (Usually false, changed to true only from test code ) If the existing test code works with this policy, it is currently OK. If a more elegant development method is discovered, it would be better to follow it. 

There is a trancecode path in the place where you have a problem like the one above. However, here there are hidden path options, which are only rewritten during testing. (Normally, the trancecode directly under the data directory is used, but when running CI, ffmpeg is installed on Github Action, and the trancecode is tested using the path installed by Github.) If a revolutionary method is found, it would be a good idea to uniformly follow it.

jpsonic-main/src/main/java/com/tesshu/jpsonic/domain/PlayQueue.java
tesshucom: Done.

airsonic-taglibs/src/main/java/com/tesshu/jpsonic/util/StringUtilBase.java
```diff
@@ -39,7 +39,6 @@
 public final class StringUtilBase {
```
tesshucom, line 39: Place the formatter in a ThreadHolder.

jpsonic-main/src/main/java/com/tesshu/jpsonic/controller/M3UController.java
```diff
@@ -89,10 +89,7 @@ private void createClientSidePlaylist(PrintWriter out, Player player, String url
             out.print("\ufeff");
         }
         out.println("#EXTM3U");
-        List<MediaFile> result;
-        synchronized (player.getPlayQueue()) {
-            result = player.getPlayQueue().getFiles();
-        }
+        List<MediaFile> result = player.getPlayQueue().getFiles();
```
tesshucom, line 92: It is OK if player#getPlayQueue()#getFiles() is synchronized

jpsonic-main/src/main/java/com/tesshu/jpsonic/controller/StatusChartController.java
```diff
@@ -184,8 +191,12 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
         rangeAxis.setTickMarkPaint(fgColor);
         rangeAxis.setAxisLinePaint(fgColor);
 
-        synchronized (LOCK) {
+        try {
```
tesshucom, line 194: JFreeChart's drawing itself is unthread safe, but since the dataset is local, I don't think there will be any problems 🙄 The past use of synchronized seems to be a misunderstanding of what causes IO errors. It's not a super important feature, so it will probably be rewritten and monitored.

jpsonic-main/src/main/java/com/tesshu/jpsonic/controller/UserChartController.java
```diff
@@ -85,7 +89,14 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
 
-        ChartUtils.writeChartAsPNG(response.getOutputStream(), chart, IMAGE_WIDTH, imageHeight);
+        try {
+            ChartUtils.writeChartAsPNG(response.getOutputStream(), chart, IMAGE_WIDTH, imageHeight);
+        } catch (IOException e) {
+            if (LOG.isTraceEnabled()) {
+                LOG.trace("Client may have closed the Stream.", e);
+            }
+        }
+
```
tesshucom, line 99: Same as StatusChartController

tesshucom:
Squash is possible. There is one strange code, so additional confirmation is required.

jpsonic-main/src/main/java/com/tesshu/jpsonic/domain/PlayQueue.java
```diff
@@ -94,24 +95,28 @@ public void setName(String name) {
...
-                if (index == -1 || index == 0 && size() == 0) {
-                    setStatus(Status.STOPPED);
-                    return null;
-                } else {
+        readLock(sequenceLock);
+        try {
+            if (index == -1 || index == 0 && size() == 0) {
+                setStatus(Status.STOPPED);
+                return null;
+            } else {
```
tesshucom, line 103: Is this code necessary or unnecessary?

tesshucom: ah,
This is strange code if we want to ensure only the atmic and the visibility, of the "List<MediaFile> files". However, there are cases where the status is changed after checking the existence of the file.

Whatever this behavior is, it may be best not to change it. But, getCurrentFile() requires a writeLock, not a readLock. (While this method is called, inquiries about its status should wait.)

tesshucom: Done.

Issue #2587, opened 2024-02-20
[UPnP Menu] Migrate to customizable menu


tesshucom:
Prerequisites: #2349

## Overview

Allow user to enable/disable menu items, customize order, and name

## Details

![image](https://github.com/tesshucom/jpsonic/assets/27724847/3fb0e669-e1f4-45c4-aa3f-a4cdd79c2fb4)

The settings that users can make can be roughly divided into two groups.

#### Edit Menu item (left side of table)

Most music apps will have something similar to these somehow. When users see these, they can somehow understand the functions. On the other hand, if what you expected isn't listed there, user get angry. Users also get angry if there are many items. Furthermore, even if all the expected items are present, if they are not displayed in the expected order, the user may feel a sense of discomfort or some sort of intent. Therefore, it is better to let the user decide these.

The image is that if you customize it on the settings screen, it will be displayed exactly as you want on the smartphone app. There is an advantage that you can shorten the Album Artist notation to Artist or change it to your native language. 

 -  Menu items like Folder, Album Artist, Album, Genre, Podcast, Playlists, Recently, Shuffle, Year are provided by default. Initially, only existing items.
 -  Menu items can be enabled/disabled, ordered, and named. 

#### Edit submenu items (right side of table)

 -  All currently existing UPnP items will be changed to these child items.
 - Children can be enabled/disabled with checkboxes
   - If only one is selected in the group, it will be used directly
   - Hierarchical display if multiple selections are made. This allows the user to control the hierarchical structure
   - For example, there are cases where it is better to display Artist as an index, and cases where it is better to display it as an enumeration instead of an index. This may depend on the UI of the smartphone app, and it also depends on the number of songs. Some people are fine with one or the other for the rest of their lives, while others want to use different ones.

#### Add Reset Buttons

 -  To avoid confusion for first-time users, a reset button will be added to make the combination as general as possible.
   - Only the most appropriate child will be enabled.
     - One of the conditions is that it should be possible to display the general information associated with the function name.
     - The other condition is that it complies with the UPnP specification.
       - For example, an index of ID3 conforms to the UPnP specification. On the other hand, File Structure indexes do not conform to the UPnP specification (UPnP strictly defines the hierarchical classes of File Structure. Hardware device/Partitions/folder, something like that.)

PR #2615, opened 2024-04-12
Migrate UPnP menu to customizable

tesshucom:
Prerequisites: #2587

## Overview

Users will have the ability to customize their UPnP content list more flexibly. 

Jpsonic's UPnP has long been a style that provides multiple views with slightly different specifications, and allows the user to enable the one of them that they want to use. This pull request will allow users to rename and reorder UPnP items.

## Goal

 - Supports enabling/disabling, renaming, and reordering of UPnP items.
 - Menu settings will be changed to be managed on the DB instead of the Property file.
 - In preparation for future feature additions. Make it easy to add missing items.

## Non-Goal

 - This pull request only rewrites the previous menu management structure. No new views have been added.
 - Enabling/disabling settings previously set by the user will not be migrated. **Please note that your previous settings will be discarded.**

## Web page changes

<details>
<summary>UPnP settings Page has been changed</summary>

"UPnP basic settings" remains unchanged. Newly added "Menu settings" and "Customizing content display".

image

</details>
<details>
<summary>Menu settings</summary>

Currently, there are eight basic menu items such as FOLDER and ARTIST. New items are planned to be added to these menu items in v114.x.

 - Can be enabled/disabled.
 - You can set the display name.
   - To avoid confusion, the initial setting is "Album Artist", but it could be abbreviated to "Artist".
   - The purpose is also to avoid i18n. Users can set it to their native language.
 - It is possible to change the order. However, it is powerless against Apps that overwrite sorting. Choose good client software. (In UPnP, Apps that force client-specific sorting, at least by default, can be called anarchy.)

Settings can be changed instantly. No service restart is required. With software such as BubbleUPnP, which has an excellent cache refresh design, it will be easy to confirm this.

image

</details>
<details>
<summary>Customizing content display</summary>

 - It is possible to change the display contents of the menu. 
 - By default, one item is enabled for each item. If multiple Items are enabled for a single menu, the UI will display more nested structures.
   - Display name will only be used if multiple items are enabled. If only one is enabled, its contents will be displayed directly. To minimize nesting of the UI, only one item should be enabled. 
   - If all items are disabled for a menu, the default item will be enabled.
   - New items will be added to these sub-items in v114.x.
   - Internally, these items can be sorted, but there are no plans to implement this functionality for the time being.
     - By design, items that will not surprise the user are provided at the top of the list. Items that currently do not have such general views may have new views added in the future..
     - Or items with lower cross-platform characteristics will not be ranked higher. (For example, items that depend on file dates are unlikely to be designed as default value.) In other words, they are not arranged completely based on the user's senses, so it may not meet your expectations.

image

</details>

## Contents of changes

 - [x] Change settings page to allow customization of UPnP menu
 - [x] Add Reset Buttons to UPnP menu settings
 - [x] Remove unnecessary Settings Properties
 - [x] Add test cases
 - [x] Add Simple Help


jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/ProcId.java
```diff
@@ -50,4 +52,26 @@ public String getValue() {
     public static ProcId of(String s) {
         return Stream.of(values()).filter(id -> id.getValue().equals(s)).findFirst().get();
     }
+
+    public static ProcId of(MenuItemId menuItemId) {
```
tesshucom, line 56: "From" is often used?

tesshucom: Done.

PR #2653, opened 2024-05-31
Fix to suppress redundant logs at startup

tesshucom:
## Ovewview

As part of the migration to Spring 3.3.0, two redundant warning logs will be recorded when Jpsonic starts. These will be temporarily suppressed as they do not affect operation.

 - [Linux] Alias ​​Checker False Positives
 - [Win/Linux] Spring Configuration Warning

These will be tracked and will likely be fixed in later versions. At that time, log suppression will also be removed.


tesshucom:
Well, this problem is related to the Migration of Spring Boot 3.3. This Pull Request will be released at the same time as the Migration, so no announcement to users is necessary.

PR #2662, opened 2024-07-24
Add upnp id3genre

tesshucom:
## Overview

The main theme of the next release, v114.2, is improving genre browsing in UPnP. This Pull request is the first in a series of improvements.

The newly implemented Genre processing in Jpsonic does not depend on the genre master used in Subsonic/Airsonic. This eliminates the mysterious genre processing that existed in previous servers. It is expected that users will be able to browse genres intuitively.

image

  - New `Genre/AlbumId3/Song` views will be added to UPnP. This is probably one of the most common view implementation.
  - The existing `Genre/Song` view will also be partially improved.

## Goal

This will enable genre filters that are more intuitive to users, and will be redesigned to allow compound filters that take into account classifications other than genre.

## Non-Goal

The new Genre processing mechanism will be introduced in advance only in UPnP. In time, this mechanism may be ported to the Web and/or the Subsonic API.

## Probrems

Inconsistencies in genre processing were already pointed out by users during the Airsonic era.

 - https://github.com/airsonic/airsonic/issues/1046

In the first place, there is no difference between Subsonic and Airsonic in the processing related to Genre. Therefore, users who often use Genre may notice its awkward operation specifications. Of course I fully understand the causes of this, but the amount of changes required to implement it to eliminate it would be enormous. A significant number of fixes had to be worked through in stages to achieve a permanent solution. And this pull request indicates that the work is nearing completion.

___

The Subsonic/Airsonic genre aggregation is poorly designed. It needs to be fixed.

There are some problems with Legacy Servers Scan:

 - Genre counting depends on the scan flow, and the results are strongly influenced by the structure of the file structure. In particular, when counting genres by album, there is a possibility that "missing" data may occur due to the placement of tag data. And, the results also depend on the order in which the files are parsed, so you may see different results on each platform.
 - The counting for albums and songs is done using an algorithm that counts up while scanning.
   - This method is highly inefficient and difficult to scale to meet user needs.
   - This will be a barrier to partial scanning. Jpsonic has decided to introduce a feature to skip analyzing specific music folders in the future. (Yes, it's important to use multi-threading to speed up scanning, but eliminating unnecessary spinning of your hard disk is more important than making it spin faster. KAIZEN)
 - Multiple genres are not taken into consideration

To solve these issues, Jpsonic has already redesigned the entire scan flow. And, there are also issues with the aggregation itself, which will be mainly addressed by this pull request.

### Approach

The Subsonic/Airsonic data base traditionally has a master table called Genre. Strictly speaking, it is more like an aggregate table, where album_count and song_count are stored. By design, this does not need to exist in the Database. Jpsonic does not use the traditional genre table to get the list of genres. Genre table is only kept for backward compatibility. 

Let's shine some light in the right direction. In Jpsonic, the list is dynamically generated using the following conditions. On the other hand, since the these conditions are not taken into account in traditional table data, it is not possible to create accurate lists, counts, or sorts.

```
public record GenreMasterCriteria(List<MusicFolder> folders, Scope scope, Sort sort, MediaType... types) {

    public enum Scope {
        ALBUM, SONG;

        public static Scope of(String s) {
            return Stream.of(values()).filter(scope -> scope.name().equals(s)).findFirst().orElse(ALBUM);
        }
    }

    public enum Sort {
        FREQUENCY, NAME, ALBUM_COUNT, SONG_COUNT;

        public static Sort of(String s) {
            return Stream.of(values()).filter(sort -> sort.name().equals(s)).findFirst().orElse(FREQUENCY);
        }
    }
}
```

As long as the aggregation process is fast enough, it possible to aggregate data to achieve accurate `Genre/Album/Song` views. Of course, it would be possible to provide arranged views such as `Music Folder/Genre/Album/Song`. (probably what users who manage large libraries would want.)

Though a bit old, a blog post from Elastic presents the idea of ​​replacing Subsonic storage with Elasticsearch.

 - [Subsonic utilise Elasticsearch comme base de données orientée documents](https://www.elastic.co/fr/blog/subsonic-utilise-elasticsearch-comme-base-de-donnees-orientee-documents)

Of course, this is just a theoretical discussion. This is because traditional databases have robust mechanisms such as transaction management and backup mechanisms. However, since Subsonic, fork servers directly derived from Subsonic have used Lucene. Of course, Jpsonic uses this for searches and such. The aggregation process addressed in this pull request is a further extension of this. Most of the significant Genre processing is done on Lucene.

### Additional notes

 - There will be only one index added to the database, and one type of index added in the Lucene implementation. This in itself does not conflict with previous Jpsonic versions, but you will need to enable Ignoretimestamp and rescan in order to update the data.
   - To automate this, **the database and index versions will each be incremented by one. However, that is not included in this pull request.**
   - There will be other Scan-related behavior in 114.2. A data schema version up will be included in a subsequent pull request.
 - A UPnP menu will be added, but it is assumed that the patch #2660 is applied.
   - Regarding storage schema changes. This means that once this pull request is merged, it will be easy to roll back the version of Jpsonic itself to 114.1.6. Be careful when rolling back to 114.1.0- 114.1.5.


jpsonic-main/src/main/java/com/tesshu/jpsonic/service/SearchService.java
```diff
@@ -209,7 +221,15 @@ List<MediaFile> getRandomSongsByArtist(Artist artist, int count, int offset, int
...
+     * Returns the number of genres in the specified Folders and Scope.
+     *
+     * @since 114.2.0
+     */
+    int getGenresCount(GenreMasterCriteria criteria);
+
+    /**
+     * Returns albums in a genre. The method for simulating the genre specification of legacy servers. Use
+     * {@link #getAlbumId3sByGenres(String, int, int, List)}, if you don't need backward compatibility.
      *
```
tesshucom, line 233: No need to change the comment.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/scanner/ScannerProcedureService.java
```diff
@@ -561,6 +567,7 @@ int updateAlbumId3s(@NonNull Instant scanDate, boolean withPodcast) {
                     Album album = albumId3Of(scanDate, folder.getId(), song, registered);
                     Optional.ofNullable(albumDao.updateAlbum(album)).ifPresent(updated -> {
                         indexManager.index(updated);
+                        mediaFileDao.updateChildrenLastUpdated(album, scanDate);
```
tesshucom, line 570: Consider a test case for updating tags.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/scanner/WritableMediaFileService.java
```diff
@@ -153,6 +154,49 @@ MediaFile getMediaFile(@NonNull Instant scanDate, @Nullable Path path) {
...
+     * If any children are deleted, set ChildrenLastUpdated of any existing children to FAR_FUTURE.
+     */
+    private void updateSongChildrenLastUpdated(@NonNull MediaFile parent) {
+        for (MediaFile child : mediaFileDao.getChildrenOf(parent.getPathString())) {
+            if (child.getMediaType() == MediaType.MUSIC || child.getMediaType() == MediaType.AUDIOBOOK
+                    || child.getMediaType() == MediaType.VIDEO) {
+                mediaFileDao.updateChildrenLastUpdated(child.getPathString(), FAR_FUTURE);
+            }
+        }
+    }
```
tesshucom, line 198: Consider a test case for updating tags.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/search/IndexManager.java
```diff
@@ -129,6 +135,11 @@ private enum GenreSort {
 
     private final Map<IndexType, SearcherManager> searchers;
     private final Map<IndexType, IndexWriter> writers;
+
+    /**
+     * @deprecated Use Ehcache
+     */
+    @Deprecated
```
tesshucom, line 142: Replace it.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/AlbumId3ByGenreProc.java
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * Jpsonic is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * (C) 2018 tesshucom
```
tesshucom, line 17: Fix comment.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/SongByGenreProc.java
```diff
@@ -34,17 +39,27 @@
...
         this.factory = factory;
         this.searchService = searchService;
     }
 
+    private GenreMasterCriteria createGenreMasterCriteria() {
+        return new GenreMasterCriteria(util.getGuestFolders(), Scope.SONG,
+                Sort.of(settingsService.getUPnPSongGenreSort()), TYPES);
+    }
+
     @Override
```
tesshucom, line 63: SongByGenreProc : The rewrite is insufficient. Rewrite it so that the previous GenreMaster is no longer used.

jpsonic-main/src/main/resources/ehcache.xml
```diff
@@ -214,8 +214,8 @@ used with JMX monitors.
            overflowToDisk="false"
            statistics="true"/>
 
-    <cache name="searchCache"
-           maxElementsInMemory="30"
+    <cache name="genreCache"
+           maxElementsInMemory="10"
            eternal="false"
            timeToLiveSeconds="10"
```
tesshucom, line 220: Consider increasing timeToLiveSeconds. Decide after measuring 100,000 songs.

jpsonic-main/src/test/resources/ehcache.xml
```diff
@@ -214,8 +214,8 @@ used with JMX monitors.
            overflowToDisk="false"
            statistics="true"/>
 
-    <cache name="searchCache"
-           maxElementsInMemory="30"
+    <cache name="genreCache"
+           maxElementsInMemory="10"
            eternal="false"
            timeToLiveSeconds="10"
```
tesshucom, line 220: timeToLiveSeconds ...

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/SearchService.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/AlbumId3ByGenreProc.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/SongByGenreProc.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/dao/AlbumDao.java
```diff
@@ -306,6 +306,20 @@ where present and folder_id in (:folders) and year between :toYear and :fromYear
...
+        Map<String, Object> args = LegacyMap.of("folders", MusicFolder.toIdList(folders), "genres", genres, "count",
+                count, "offset", offset);
+        return template.namedQuery("select " + QUERY_COLUMNS + """
+                from album
+                where present and folder_id in (:folders) and genre in (:genres)
+                order by album_order
+                limit :count offset :offset
+                """, rowMapper, args);
+    }
+
```
tesshucom, line 322: Speed ​​Considerations
1) Consider adding an index to the genre field. If it has a negative effect, it will not be added.
2) Move to com.tesshu.jpsonic.dao.dialect.
 - Searching with Lucene is faster. (Only id searches are used in the database. However, when using the in clause, overflow may occur depending on the use of the database, so it is assumed that unnest is used. In any case, the database is limited, so it is treated as dialect.)

jpsonic-main/src/main/java/com/tesshu/jpsonic/dao/MediaFileDao.java
```diff
@@ -727,12 +784,18 @@ public List<MediaFile> getSongsByGenre(final List<String> genres, final int offs
                 Arrays.asList(MediaType.MUSIC.name(), MediaType.PODCAST.name(), MediaType.AUDIOBOOK.name()), "genres",
                 genres, "count", count, "offset", offset, "folders", MusicFolder.toPathList(musicFolders));
         return template.namedQuery("select " + prefix(QUERY_COLUMNS, "s") + """
+                , al.media_file_order = -1 as is_under_root
                 from media_file s
                 join media_file al
-                on s.parent_path = al.path join media_file ar on al.parent_path = ar.path
+                on s.parent_path = al.path
+                left join media_file ar
```
tesshucom, line 791: Move to com.tesshu.jpsonic.dao.dialect. Similar reasons to AlbumDao.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/scanner/WritableMediaFileService.java
```diff
@@ -32,6 +32,7 @@
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
+import java.util.Iterator;
```
tesshucom, line 35: Unnecessary

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/scanner/ScannerProcedureService.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/scanner/WritableMediaFileService.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/scanner/WritableMediaFileService.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/dao/MediaFileDao.java

tesshucom:
Speed ​​Considerations

- Consider adding an index to the genre field. If it has a negative effect, it will not be added.

jpsonic-main/src/main/java/com/tesshu/jpsonic/dao/AlbumDao.java
tesshucom: Done. Adding an index will be considered during speed verification.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/search/IndexManager.java
tesshucom: Done.

jpsonic-main/src/test/java/com/tesshu/jpsonic/service/scanner/MediaScannerServiceImplTest.java
```diff
@@ -516,6 +525,163 @@ void testUpdateAlbum() {
...
+            assertEquals("GENRE_G", genres.get(8).getName());
+            assertEquals(1, genres.get(8).getAlbumCount());
+            assertEquals("GENRE_H", genres.get(9).getName());
+            assertEquals(1, genres.get(9).getAlbumCount());
+            assertEquals("GENRE_K", genres.get(10).getName());
+            assertEquals(2, genres.get(10).getAlbumCount());
+
+            assertEquals("GENRE_L", genres.get(11).getName());
+            assertEquals(2, genres.get(10).getAlbumCount()); // (2 -x-> 1)
+
```
tesshucom, line 678: Test case seems to be incorrect! In the correct case it will probably work as expected.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/SongByGenreProc.java
```diff
@@ -34,17 +39,27 @@
...
         this.util = util;
         this.factory = factory;
         this.searchService = searchService;
     }
 
+    private GenreMasterCriteria createGenreMasterCriteria() {
+        return new GenreMasterCriteria(util.getGuestFolders(), Scope.SONG,
+                Sort.of(settingsService.getUPnPSongGenreSort()), TYPES);
+    }
+
```
tesshucom, line 62: There should be no problem rewriting SongByGenreProc to make it compatible with the new GenreMaster. However, AlbumByGenreProc (Subsonic-style browsing) will be affected because it inherited SongByGenreProc. By removing the inheritance of AlbumByGenreProc and transferring the processing related to GenreMaster that was previously done in SongByGenreProc to AlbumByGenreProc, the same behavior as before will be maintained.

jpsonic-main/src/test/java/com/tesshu/jpsonic/service/scanner/MediaScannerServiceImplTest.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/upnp/processor/SongByGenreProc.java
tesshucom: Done.

jpsonic-main/src/main/resources/ehcache.xml
tesshucom: Done.

jpsonic-main/src/main/resources/liquibase/jp114.2.0/add-upnp-album_id3_by_genre.xml
tesshucom, line 1: Change file names to hyphens instead of underscores.

tesshucom:
About whether to add an index to album#genre.


#### Scan

This was a bit unexpected, but adding the index made it about 1 second faster. I guess the benefit of search is greater than the load of registration...

Therefore, it is corrected as follows:

> Compared to the latest version (5 trials each): 
> 
>  - First Full Scan : an increase of about ~23 seconds~ -> 22 seconds
>  - Ignore Timestamp : Almost the same as First Full Scan
>  - Subsequent scans : Reduced by 1 to 2 seconds


The bottom line is that this is pretty close to the impact originally predicted.

We can easily surpass this later on. However, some scanning fixes are expected, such as those included in this pull request. `Premature optimization is the root of all evil.`  Right now, what's most important is making it work as accurately and robustly as possible!

#### Browsing

Almost no change. This was expected.

There is room for optimization in Browsing, but only if you assume a fairly large library. For libraries with tens of thousands of entries, the benefits of this optimization effort are limited. Focusing on other things now would provide more useful functionality for users.

> About whether to add an index to album#genre.

It's good to add.

jpsonic-main/src/main/resources/liquibase/jp114.2.0/add-upnp-album_id3_by_genre.xml
tesshucom: Done.

tesshucom:
The expected fix has been completed, so `SQUASH`.

PR #2672, opened 2024-08-07
Add UPnP menu: Filtering genre by folders

tesshucom:
## Overview

Two new views will be added that allow you to filter genres by music folder. This means that folder filtering is supported in the FOLDER, ARTIST, ALBUM, and GENRE menus.

image

## Additional Notes

This is a basic design consideration, but... Even if we covered all the music tags in the world, we still wouldn't be able to create a personal streaming server that was easy for the average user to use. Because music tags are a way to mark up information about the music.

To put it more simply, music tags never include other than song information. Therefore, in Subsonic and many other media servers, the file structure and tags have a complementary relationship. For example, in a search system, a lack of search results is often treated as a design bug, but since it cannot be solved by tags alone, it is necessary to make it possible to browse the file structure as well. Unless it's an unrealistic assumption that all users will maintain tags. And in particular, when file transfer or management is performed across system boundaries, such as with backups or FTP, file structure will take precedence over tags.

The Music Folder is, needless to say, the starting point for the scan. This specification is used by many software programs because it allows multiple different devices to be treated equally, but it is generally used for other purposes as well. You might have separate libraries for your wife, for your kids, or even other categories. These are partitions of your library, so to speak. Don't worry, Jpsonic has been redesigned to handle dozens of Music Folders.

___

So, Jpsonic's UPnP has improved filtering by music folder, but this will also be applied to menus other than Genre. RECENTLY, YEAR, etc. Of course this doesn't mean that it's necessary to split your Music Folder. If you want to handle your libraries with a single root, there's no problem leaving it as it is.


tesshucom:
Fix existing test cases. The test content is fine, but Ubuntu often fails when copying files(to reproduce tag edits) during testing. It is not reproduced on Windows. This part will be rewritten so that it does not use NIO2. There will be no impact on the existing processing of Jpsonic itself.

PR #2674, opened 2024-08-10
Change to be able to use audiobooks with UPnP 

tesshucom:
Prerequisites: #2662

## Overview

The UPnP features up to Jpsonic 114.1 is mainly for Music. From 114.2 onwards, it will be improved so that Audiobook can be used.

## Goal

It will add features to make it easier to browse Audiobooks with UPnP without making major changes to the Subsonic-like management method.

## Non Goal

This does not include design changes such as additional scanning of audiobook tags. This is simply an advanced features that is an extension of Subsonic/Airsonic. Of course, we're not ruling out future tag expansion, but please be aware that it's not a current high-priority task that should be prioritized. It's not that difficult. In other words, it doesn't have to be now.

## Details

There are some changes included. Also, the prerequisites should be made clear.

### Prerequisites

Subsonic servers have a category called MediaType for each file they manage. Apps need to change their behavior depending on the file. For example, if it's a music file (MUSIC), it should play music, if it's a video file (VIDEO), it should play video, etc.

A MediaType called AUDIOBOOK already exists (it is not an extension of Jpsonic). This is conditional on the Path string or the Genre string containing the string `audiobook`. This is used for shuffle exclusion in Legacy servers.

### Fixes

Based on Prerequisites.

#### Change existing UPnP views so that Audiobook can be used

Audiobooks were intentionally excluded from 114.1. This will be removed and Audiobook viewing will become available.

Why was it excluded?

> We were considering splitting the UPnP tree by MediaType if possible. Currently, the plan is for MUSIC and AUDIOBOOK to coexist. Of course, this could change in the future, but the current way of doing things also has its merits. For users who are familiar with Subsonic, the new features will be easy to accept because they can use them without changing the operational specifications. 

Conversely, filters for split browsing have been enhanced so that MUSIC and AUDIOBOOK can be used in the same tree. In reality, the most practical solution for AUDIOBOOK is to include `audiobook` in the Music Folder directory name. For this reason, most browsing on Jpsonic allows filtering by Music Folder.

#### Add a dedicated Audiobook browsing tree to the Genre submenu

A new submenu will be added. Please note that this is aimed at splitting MUSIC and AUDIOBOOK, just like excluding Shuffle.

image

Why should MUSIC and AUDIOBOOK be separated?

> Even if the same tag field `GENRE` is used, `GENRE` in MUSIC and AUDIOBOOK are from different worlds.
> - Biography & Memoir
> - Kids
> - Mystery & Suspense
> - Fiction & Literature
> - Nonfiction
> - Romance
> - Anime
> If the MUSIC and AUDIOBOOK genres are mixed, confusion or outright conflict will occur (in this example, Anime etc. would probably be mixed up, right?).

Jpsonic's Genre can display different things for each folder or MediaType. So this is possible.

#### Add a comment to production.synology.yml for starting Docker

This is not a feature fix, just a quick announcement.

>    If you use Audiobook,
>    we recommend that you include the audiobook string in the entry path.
>    In Subsonic's direct lineage servers, it is used for shuffle exclusion, etc.
>    In Jpsonic, you can separate music and audiobook genres (with UPnP).
>    
>    - '/volume1/SomeDir:/jpsonic/audiobook'

Docker allows path mapping, making it very easy to use. In the above example, `/jpsonic/audiobook` contains the word audiobook, so your local directory can have any name.

Of course, just as you can have multiple Music Folders for music, you can have multiple directories for Audiobooks.


install/docker/production.synology.yml
```diff
@@ -44,6 +44,13 @@ services:
     # - '/volume1/Music2:/jpsonic/music2'
     # - '/volume1/Music3:/jpsonic/music3'
 
+    # If you use Audiobook,
+    # we recommend that you include the audiobook string in the entry path.
+    #　In Subsonic's direct lineage servers, it is used for shuffle exclusion, etc.
```
tesshucom, line 49: Full-width space

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/search/IndexManager.java
```diff
@@ -741,23 +739,23 @@ private List<GenreFreq> getGenreFreqs() {
...
-            Query query = queryFactory.getSongGenreCount(genreName, criteria.folders(), types);
-            return searcher.search(query, new TotalHitCountCollectorManager());
+    private int getSongGenreCount(IndexSearcher searcher, String genreName, GenreMasterCriteria criteria) {
+        try {
+            MediaType[] types = criteria.types().length == 0 ? MUSIC_AND_AUDIOBOOK : criteria.types();
+            return searcher.search(queryFactory.getSongGenreCount(genreName, criteria.folders(), types),
+                    new TotalHitCountCollectorManager());
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
         }
```
tesshucom, line 758: To be precise, this fix increases the genre logic according to criteria. However, it is not necessary to reproduce this exactly in your coding. This logic was written to avoid unnecessary counting. However, the counting speed was much faster than expected (fast enough that the extra avoidance was not necessary). So in this commit, they are rather simplified.

Just to be sure, Check the speed using data of 100,000 songs and 100 genres or more.

tesshucom: Done.

install/docker/production.synology.yml
tesshucom: Done.

Issue #2691, opened 2024-09-15
Do not expose song information in Subsonic API's Nowplaying


tesshucom:
Prerequisites: #2688

___

## Overview

We need to set it so that song information is not made public in NowPlaying of API.

Currently, the default setting for Jpsonic is to not let others know what song you are playing on NowPlaying. This is not just limited to web pages; similar restrictions are also in place for the Subsonic API.

Is it because you are embarrassed to make your preferences known? It's not. **This is based on reasonable design standards that place restrictions on the ability to actively disclose account names to others.**

 - This is for the same reason that logging and network information is only accessible to privileged users. Security requirements take precedence over Usability you are imaging.
 - We often see a feature in apps such as social media that allows you to set a display name that is different from your account name. In the case of a system that considers account names as semi-personal information, such a design would likely be adopted. There is no such field in the Subsonic API, nor do I have any desire to add one.

## Fixes

Currently, individuals can choose whether or not to make information about their playing song public. (The setup is very complicated.) This will be deleted once.

Instead, it would be appropriate to change the settings to those set by privileged users on the following User Settings Page. User individually configurable features will be omitted, and only privileged users will be able to grant permissions.

![image](https://github.com/user-attachments/assets/5eacc2c3-e874-455f-814d-fb2b288bd816)

PR #2692, opened 2024-09-16
Remove or modify unnecessary features

tesshucom:
Prerequisites: #2585

## Overview

Features that were previously considered for removal will either be removed or have their specifications changed. See [Suppressed or Removed Features](https://github.com/jpsonic/jpsonic/wiki/Suppressed-or-Removed-Features).

The target features are as follows

 - [x] Publish server log to general users
 - [x] Publish network usage status to general users
 - [x] Use list of songs being played
 - [x] Publish playlist on Posdcast
 - [x] Use external player
 - [x] Copy non-ascii name files when transcoding
 - [x] Remove track number from ID3 song title
 - [x] Use Clean-up
 - [x] Do verbose folder checks
 - [ ] Fix Wiki

## Details

Any issues I noticed while adding a feature or reviewing the source code were suppressed as a reminder. Therefore, it can be said that this is an issue that has been put off for a while.

I think it might fall into one of the following categories, there are based on general principles:

 - It is undesirable for non-privileged users to easily learn other account names or their activities.
 - Servers must not automatically collect or disclose information through features unknown to users.
 - An implementation bug that appears to be working around some problem, but is actually causing other serious problems

Features such as a list of songs currently playing from other users are likely to be in demand by "some users." However, it is a fact that no engineer has come forward to make any improvements even after two generations of Subsonic and Airsonic. This suggests that if left unchecked, it is likely to remain so for generations to come. By the way, in my personal opinion, this is not a critical feature and should be removed unless it is a valid design. By saving time on that improvement, you can allocate resources to other, more beneficial improvements. Life is finite, and the development of freeware still requires significant resources and effort.

## Goal

Already suppressed features will be removed, i.e. the goal is to completely preserve the current default behavior while removing redundant source code.

## Non-Goal

The web page implementation that lets you see what songs others are listening to will be removed. The ugly timer communication is being removed. Hooray!

However, please note that this functionality also exists in the Subsonic API. In other words, this function is not completely rejected. If you use an app that already implements that functionality via an API, it will be available to you.

Currently, Jpsonic has restrictions in place on the API as well, and these will need to be maintained. However, this pull request does not include that fix, it will be implemented in a subsequent pull request. Related #2691.


jpsonic-main/src/main/java/com/tesshu/jpsonic/ajax/PlayQueueService.java
tesshucom, line 1: Add test cases.

jpsonic-main/src/main/webapp/WEB-INF/jsp/playerSettings.jsp
```diff
@@ -152,20 +132,6 @@
...
-                            <form:checkbox path="autoControlEnabled" id="autoControlEnabled" cssClass="checkbox"/>
-                            <label for="autoControlEnabled"><fmt:message key="playersettings.autocontrol"/></label>
-                            <c:import url="helpToolTip.jsp"><c:param name="topic" value="autocontrol"/></c:import>
-                        </dd>
-                        <dt></dt>
-                        <dd>
-                            <form:checkbox path="m3uBomEnabled" id="m3uBomEnabled" cssClass="checkbox"/>
-                            <label for="m3uBomEnabled"><fmt:message key="playersettings.m3ubom"/></label>
-                            <c:import url="helpToolTip.jsp"><c:param name="topic" value="m3ubom"/></c:import>
-                        </dd>
```
tesshucom, line 167: Delete messages from a property file.

jpsonic-main/src/main/webapp/WEB-INF/jsp/personalSettings.jsp
```diff
@@ -519,20 +517,6 @@ document.addEventListener('DOMContentLoaded', function () {
...
-                <c:import url="helpToolTip.jsp"><c:param name="topic" value="nowplayingallowed"/></c:import>
-            </dd>
-            <c:if test="${command.othersPlayingEnabled}">
-                <dt><fmt:message key="personalsettings.menu"/></dt>
-                <dd>
-                    <form:checkbox path="showNowPlayingEnabled" id="nowPlaying" />
-                    <label for="nowPlaying"><fmt:message key="personalsettings.shownowplaying"/></label>
-                    <c:import url="helpToolTip.jsp"><c:param name="topic" value="shownowplaying"/></c:import>
-                </dd>
-            </c:if>
```
tesshucom, line 535: Delete messages from a property file.

jpsonic-main/src/main/webapp/WEB-INF/jsp/musicFolderSettings.jsp
tesshucom, line 1: Line feed code.

jpsonic-main/src/main/webapp/WEB-INF/jsp/dialogs.jsp
```diff
@@ -24,8 +24,6 @@
     </form>
 </div>
 
-<div id="dialog-nowplayinginfos" title="<fmt:message key='top.othersplaying'/>"></div>
```
tesshucom, line 27: Delete messages from a property file.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/PlayerService.java
```diff
@@ -96,9 +93,6 @@ public void init() {
         writeLock(playerLock);
         try {
             playerDao.deleteOldPlayers(60);
-            if (!settingsService.isUseExternalPlayer()) {
-                resetExternalPlayer();
```
tesshucom, line 100: ExternalPlayer is already "OldPlayer" so it will need to be deleted regardless of its expiration date.

jpsonic-main/src/main/java/com/tesshu/jpsonic/controller/SubsonicRESTController.java
```diff
@@ -1326,11 +1325,6 @@ public void getNowPlaying(HttpServletRequest req, HttpServletResponse response)
                 continue;
             }
 
-            UserSettings userSettings = securityService.getUserSettings(username);
-            if (!userSettings.isNowPlayingAllowed()) {
-                continue;
-            }
-
             long minutesAgo = status.getMinutesAgo();
             if (minutesAgo < LIMIT_OF_HISTORY_TO_BE_PRESENTED) {
```
tesshucom, line 1329: This code will be redesigned. Please confirm.

jpsonic-main/src/main/java/com/tesshu/jpsonic/controller/MoreController.java
```diff
@@ -77,13 +73,11 @@ protected ModelAndView handleRequestInternal(HttpServletRequest request, HttpSer
...
         }
 
-        Player player = playerService.getPlayer(request, response);
         ModelAndView result = new ModelAndView();
         result.addObject("model",
                 LegacyMap.of("user", user, "uploadDirectory", uploadDirectory, "genres", searchService.getGenres(false),
                         "currentYear", Calendar.getInstance().get(Calendar.YEAR), "musicFolders", musicFolders,
-                        "clientSidePlaylist", player.isExternalWithPlaylist() || player.isWeb(), "brand",
-                        SettingsService.getBrand()));
+                        "clientSidePlaylist", false, "brand", SettingsService.getBrand()));
```
tesshucom, line 80: Isn't clientSidePlaylist unnecessary?

jpsonic-main/src/main/java/com/tesshu/jpsonic/controller/M3UController.java
tesshucom, line 1: This class had a lot of unnecessary code... So no test code was written. For the remaining code, check the callers. Of course, if it is not needed, it should be deleted.

jpsonic-main/src/main/java/com/tesshu/jpsonic/controller/SubsonicRESTController.java
tesshucom: Although it will be removed in this commit, similar controls will likely be added back in a subsequent pull request.



 - userSettings#NowPlayingAllowed will not be removed from the Database schema and will be reused. As shown in the code above, it is used in API controllers to control information disclosure.
 - The userSettings#NowPlayingAllowed setting is added to the user Settings page with a default of false.
   - #2691
 - By enabling the flag for the account that the Admin wants to make public, Nowplaying will be available just like regular Subsonic. 
   - This is not something that is decided by the user's preferences, but by the server administrator, taking into account the operational policy (just like other user settings). The main concerns for server administrators are server load and rules on how account name disclosure should be handled.

jpsonic-main/src/main/java/com/tesshu/jpsonic/ajax/PlayQueueService.java
tesshucom: Podcasts and Internet Radio may contain some questionable code. These have not yet been fully implemented in Jpsonic, but will be reimplemented in a later version.

jpsonic-main/src/main/webapp/WEB-INF/jsp/playerSettings.jsp
tesshucom: Done.

jpsonic-main/src/main/webapp/WEB-INF/jsp/personalSettings.jsp
tesshucom: Done.

jpsonic-main/src/main/webapp/WEB-INF/jsp/dialogs.jsp
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/controller/MoreController.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/scanner/ScannerProcedureService.java
```diff
@@ -219,9 +219,7 @@ void beforeScan(@NonNull Instant scanDate) {
         mediaFileCache.setEnabled(false);
         mediaFileCache.removeAll();
 
-        if (!settingsService.isUseCleanUp() && mediaFileDao.existsNonPresent()) {
-            // If useCleanUp=false(default), Clean-up can be managed by Scan processes.
-            // Removing present=false in advance can reduce the number of subsequent queries issued.
+        if (mediaFileDao.existsNonPresent()) {
```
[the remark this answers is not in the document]
tesshucom: Yes, it's unnecessary, but it's only a part of the whole unnecessary thing. It would be better to have it deleted in a planned manner.

Subsonic/Airsonic uses logical deletion, but the drawbacks far outweigh the benefits and it was the cause of many many many many problems. Therefore, Jpsonic has been redesigned to not use logical deletion. So it could be removed, but it's probably best to defer this to a later version.

 - Assuming that the option to use Clean Up is enabled, it would be better for "Delete logical-deletion" to be performed at least in a version later than the Clean Up version.
 - This code isn't the only legacy of logical deletion. There is other trash too.
   - First, present-flag should be removed from all SQL.
   - There is still some crappy code, like the Ajax PlaylistService#populateAccess. (This is not a logical deletion, but is used for other purposes filter.) It will need to be rewritten correctly.
   - We will need to remove this code as well as any clean up code that assumes logical deletion.

Doing this will eliminate the side effects. Maintain status quo.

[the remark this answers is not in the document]
tesshucom: It would be better if a new issue was issued regarding this matter.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/PlayerService.java
tesshucom: Done.

jpsonic-main/src/main/java/com/tesshu/jpsonic/service/scanner/ScannerProcedureService.java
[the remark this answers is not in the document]
tesshucom: #2693

jpsonic-main/src/main/java/com/tesshu/jpsonic/controller/M3UController.java
tesshucom: The controller should be removed as it is not needed. By removing this, the sendM3U flag will no longer be necessary, which should simplify your Ajax processing a bit. 

Although it is the same m3u, ExportPlayListController is used to save the Playlist, which has the extension m3u8 and the processing is slightly different. M3UController is a Controller for exporting m3u files directly to external players.

tesshucom: Done.

tesshucom:
It seems likely that the original planned modifications have been covered. The scope of the impact is large, so it will be squash merge after the actual operation is confirmed.

